PR #11, opened 2018-02-15
sql/parse: use sqlparser to handle DESCRIBE queries

### LINKED ISSUE #5: Modify DESCRIBE to use the sql parser
Instead of use a regular expression: 
https://github.com/src-d/go-mysql-server/blob/50166d8f9e431566f0db464f77cc710a92e0a8b3/sql/parse/parse.go#L37

Modify the code to use the standard query to describe the fields of a specific table.

erizocosmico:
Closes #5 
Depends on https://github.com/src-d/go-vitess/pull/2

### DIFF
--- a/sql/parse/parse.go
+++ b/sql/parse/parse.go
@@ -1,8 +1,8 @@
 package parse
 
 import (
+	"errors"
 	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


erizocosmico:
Obsolete, will use `SHOW COLUMNS FROM <table>`

PR #12, opened 2018-02-15
Fix tests and CI

ajnavarro:


.travis.yml
```diff
@@ -1,20 +1,21 @@
 language: go
-sudo: false
+
 go:
-  - 1.7
-  - 1.8
+  - 1.9
```
mcarmonaa: better if we use 1.9.x?

mcarmonaa: nevermind, it's the same https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use :+1:

erizocosmico: 1.9 is the latest release in the 1.9 minor version (that is, 1.9.x), so it's actually the same

PR #13, opened 2018-02-15
*: implement distinct clause

### LINKED ISSUE #2: Add DISTINCT aggregation

erizocosmico:
Closes #2 

This PR implements the DISTINCT SQL Node. Distinct wraps a projection and emits only rows that have not been previously seen before.

There is a common table with more than 100 rows to perform benchmarks. A Benchmark has been added for both Project and Distinct, which shows that Distinct is two orders of magnitude slower than Project. Although this is just a first approximation to have the feature, in the feature we should spend some time thinking on how to properly and efficiently implement the distinct and/or try to identify if the projection is sorted, which would eliminate the need to keep track of the hashes and make this way faster.

### DIFF
--- a/mem/table.go
+++ b/mem/table.go
@@ -55,7 +55,7 @@ func (t *Table) Insert(row sql.Row) error {
 	for idx, value := range row {
 		c := t.schema[idx]
 		if !c.Check(value) {
-			return sql.ErrInvalidType
+			return sql.ErrInvalidType.New(value)
 		}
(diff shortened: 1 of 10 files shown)


mcarmonaa:
@erizocosmico Should we open an issue about the more efficient implementation to keep it in mind?

erizocosmico:
Opened #22 to keep track of it.

PR #18, opened 2018-02-15
sql: implement offset node

erizocosmico:
Implements the offset node as a separate node


erizocosmico:
Depends on #19 so tests pass

PR #20, opened 2018-02-15
*: add comments to all exported members

erizocosmico:
Also made `golint`, `go vet` and `gofmt -s` happy and simplified some code.

### DIFF
--- a/sql/expression/aggregation.go
+++ b/sql/expression/aggregation.go
@@ -6,26 +6,32 @@ import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
 
+// Count node to count how many rows are in the result set.
 type Count struct {
 	UnaryExpression
 }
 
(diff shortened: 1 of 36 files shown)


erizocosmico:
Done, can be reviewed now

sql/expression/aggregation.go
```diff
@@ -34,15 +40,18 @@ func (c *Count) Resolved() bool {
 	return c.Child.Resolved()
 }
 
+// Name returns the of the node.
```
mcarmonaa: s/returns the of/returns the name of

erizocosmico:
Depends on #19 so tests pass

sql/expression/aggregation.go
erizocosmico: good catch

erizocosmico:
Fixed @mcarmonaa

PR #26, opened 2018-02-16
sql/expression: mark as placeholder unresolved expressions

erizocosmico:
`UnresolvedFunction` and `UnresolvedColumn` are placeholders, instead of potentially return wrong data it's preferable to panic and let the user know these methods should not be called.

This introduced an expected bug, which is that tests fail because types of the comparison children are obtained ahead of time. So I've fixed this moving that to Eval time.


erizocosmico:
Don't review yet, have to fix the part that calls `Type` on unresolved nodes

erizocosmico:
Fixed, can be reviewed.

PR #31, opened 2018-02-16
sql: make Eval return an error as well 

erizocosmico:


sql/expression/comparison.go
```diff
@@ -68,26 +76,33 @@ func NewRegexp(left sql.Expression, right sql.Expression) *Regexp {
...
 
 	if !okl || !okr {
-		return re.ChildType.Compare(l, r) == 0
+		return re.ChildType.Compare(l, r) == 0, nil
 	}
 
 	reg, err := regexp.Compile(sr)
 	if err != nil {
-		return false
+		return false, nil
```
jfontan: should return err

erizocosmico: yes, good catch!

erizocosmico:
Fixed @jfontan

sql/plan/filter.go
```diff
@@ -55,10 +55,16 @@ type filterIter struct {
...
+			return nil, err
+		}
 
+		ok, err := i.f.expression.Eval(row)
 		if err != nil {
 			return nil, err
-		} else if i.f.expression.Eval(row) == true {
+		}
+
+		if ok == true {
```
ajnavarro: with `if ok {` I think the result is the same :)

erizocosmico: Nope: https://play.golang.org/p/XP_hLd18CJF

erizocosmico:
Rebased

sql/plan/filter.go
erizocosmico: Nope: https://play.golang.org/p/XP_hLd18CJF

ajnavarro: So, at least for me, have a `ok` variable name that usually means boolean feels a little bit strange for me. Maybe we should change the variable name for somethig less common, like `evaluated` ?

erizocosmico: Yes, that makes sense. `ok` might induce to think it's a bool

erizocosmico:
Fixed @ajnavarro

PR #49, opened 2018-02-26
sql/parse: support more than 2 tables on cross join

### LINKED ISSUE #38: CROSS JOIN doesn't support more than 2 tables
Add support for more than two tables on join.

erizocosmico:
Fix #38 

Also, this is not inner join as said in the issue, is cross join, because INNER JOIN is not implemented right now.

### DIFF
--- a/sql/parse/parse.go
+++ b/sql/parse/parse.go
@@ -193,12 +193,12 @@ func tableExprsToTable(te sqlparser.TableExprs) (sql.Node, error) {
 		return nodes[0], nil
 	}
 
-	if len(nodes) == 2 {
-		return plan.NewCrossJoin(nodes[0], nodes[1]), nil
+	join := plan.NewCrossJoin(nodes[0], nodes[1])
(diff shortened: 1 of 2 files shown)


sql/parse/parse.go
```diff
@@ -193,12 +193,12 @@ func tableExprsToTable(te sqlparser.TableExprs) (sql.Node, error) {
 		return nodes[0], nil
 	}
 
-	if len(nodes) == 2 {
-		return plan.NewCrossJoin(nodes[0], nodes[1]), nil
+	join := plan.NewCrossJoin(nodes[0], nodes[1])
```
ajnavarro: this line can be ommited?

ajnavarro: should we check the len of nodes before? (or use left, right)

ajnavarro: nevermind: https://github.com/src-d/go-mysql-server/pull/49/files#diff-913eb5b14dd573d980e4f89476047ab6R192

PR #51, opened 2018-02-26
Fix GroupBy

mcarmonaa:
- Use of `First` aggregation removed from `GroupBy`
- Rule added to validate `group by` statements.

### DIFF
--- a/sql/analyzer/validation_rules.go
+++ b/sql/analyzer/validation_rules.go
@@ -13,7 +13,7 @@ const (
 )
 
 var (
-	ValidationResolvedErr = errors.NewKind("plan is not resolved because of node the '%T'")
+	ValidationResolvedErr = errors.NewKind("plan is not resolved because of node '%T'")
 	ValidationOrderByErr  = errors.NewKind("OrderBy does not support aggregation expressions")
 	ValidationGroupByErr  = errors.NewKind("GroupBy aggregate expression '%v' doesn't appear in the grouping columns")
 )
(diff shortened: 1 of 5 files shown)


mcarmonaa:
the failing comes from the parser is parsing `SELECT COUNT(*) AS c FROM mytable;` as a `GroupBy` node, triying to fix it...

sql/analyzer/validation_rules.go
```diff
@@ -1,36 +1,86 @@
...
 )
 
+const (
+	validateResolvedRule = "validate_resolved"
+	validateOrderByRule  = "validate_order_by"
+	validateGroupByRule  = "validate_group_by"
+)
+
+var (
+	ValidationResolvedErr = errors.NewKind("plan is not resolved because of node the '%T'")
```
jfontan: I don't understand this message

sql/analyzer/validation_rules.go
```diff
@@ -1,36 +1,86 @@
...
 	}
 
 	return nil
 }
+
+func validateGroupBy(n sql.Node) error {
+	switch n := n.(type) {
+	case *plan.GroupBy:
+		// Allow the parser use the GroupBy node to eval the aggregation functions
+		// for sql statementes that aren't really make use of the GROUP BY expression.
```
jfontan: aren't really -> don't

erizocosmico:
Rebase needed

PR #55, opened 2018-02-26
sql/expression: implement substring function

### LINKED ISSUE #42: Implement SUBSTRING(string, start, length) function

erizocosmico:
Fixes #42

This PR implements the substring(str, start[, length]) function with the same rules as its homonym function in MySQL.
Since Go strings are UTF8, this function does not return a direct substring `str[start:start+length]`, instead returns the substring of runes. That is, `"á"[0:1]` does not return a partial unicode glyph, but `"á"` itself.

### DIFF
--- a/sql/expression/function.go
+++ b/sql/expression/function.go
@@ -2,6 +2,7 @@ package expression
 
 import (
 	"bytes"
+	"reflect"
 
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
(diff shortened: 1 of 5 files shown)


sql/expression/function.go
```diff
@@ -65,3 +67,107 @@ func isBinary(data []byte) bool {
...
+
+	length := ln.(int64)
+	startIdx := start.(int64)
+	runeCount := int64(len(text))
+
+	if length < 0 {
+		return nil, fmt.Errorf("substring: invalid length %d", length)
+	}
+
+	if startIdx < 0 {
```
ajnavarro: Take into account that:
- The first position in string is 1.
- If start_position is a positive number, then the SUBSTRING function starts from the beginning of the string.
- If start_position is a negative number, then the SUBSTRING function starts from the end of the string and counts backwards. Negative values for start_position was introduced in MySQL 4.1.
- If start_position is bigger than the string length, the result is NULL
- If start_position is 0, the result is NULL

erizocosmico:
Done @ajnavarro

sql/expression/function.go
```diff
@@ -65,3 +67,107 @@ func isBinary(data []byte) bool {
...
+	ln, err = sql.Int64.Convert(ln)
+	if err != nil {
+		return nil, err
+	}
+
+	length := ln.(int64)
+	startIdx := start.(int64)
+	runeCount := int64(len(text))
+
+	if length < 0 {
```
ajnavarro: In mysql substring implementation length is optional

ajnavarro: Also, If length is bigger than the remained characters, no error happens

ajnavarro:
@erizocosmico You can check all corner cases here: https://www.w3schools.com/sql/trymysql.asp?filename=trysql_func_mysql_substring

Do you mind to add test for them?

erizocosmico:
Yes, sure

sql/expression/function.go
ajnavarro: I think the UI is wrong. Check with: `SELECT SUBSTRING("SQL Tutorial", 5, 3) AS ExtractString, NULL as test;` the output of `test` appears to be "" but it is NULL

erizocosmico: If you do `SELECT NULL is null` returns `1`, and if you do  `SELECT SUBSTRING("fooo bar", 0) is null` it returns `0`, so in this case is not null what it returns. From what I see in that UI, the only way SUBSTRING returns null is if the string itself is null.

erizocosmico: https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_substring

> If len is less than 1, the result is the empty string.

erizocosmico:
Fixed @ajnavarro

PR #64, opened 2018-02-27
sql: implement Tuple type and support all literals

### LINKED ISSUE #60: Add support to sqlparser.ValTuple and sqlparser.SQLVal

erizocosmico:
Fixes #60


sql/parse/parse.go
```diff
@@ -426,9 +413,63 @@ func exprToExpression(e sqlparser.Expr) (sql.Expression, error) {
...
+			return nil, err
+		}
+		return expression.NewLiteral(val, sql.Blob), nil
+	case sqlparser.ValArg:
+		return expression.NewLiteral(string(v.Val), sql.Text), nil
+	case sqlparser.BitVal:
+		return expression.NewLiteral(v.Val[0] == '1', sql.Boolean), nil
+	}
+
+	panic(fmt.Errorf("unreachable: invalid SQLVal of type: %d", v.Type))
```
ajnavarro: we should return the error instead of panic here.

sql/type.go
```diff
@@ -375,6 +378,24 @@ func (t jsonT) Compare(a interface{}, b interface{}) int {
...
+func (t tupleT) SQL(v interface{}) sqltypes.Value {
+	panic("unable to convert tuple type to SQL")
+}
+
+func (t tupleT) Convert(v interface{}) (interface{}, error) {
+	return nil, fmt.Errorf("unable to convert tuple type")
+}
+
+func (t tupleT) Compare(a, b interface{}) int {
+	panic("unable to compare two tuple values")
```
ajnavarro: Tuples are comparable, right? `SELECT (3,4) IN ((1,2), (3,5),(3,4)) as b;`

sql/parse/parse.go
erizocosmico: Can change it to the error return, but the panic was intentional because that part should not be reachable

sql/type.go
erizocosmico: We don't have the concrete types to perform the comparison here. It could be done, but some things need to change a bit for that.
- Tuple type would need to be a Type constructor instead, to give it the concrete types of the elements in the tuple.
- `Compare` needs to change to also return an error because there is a lot of cases in which a comparison between tuples can fail (having different number of elements, for example).

jfontan: Do we need comparison now? If not we can just open an issue to add it in the future.

erizocosmico: It's something that's expected of tuples. If not, they're not very useful. Right now, since we're not using tuples yet, I think this can wait until we do and then do the aforementioned changes, which are vital to implement this correctly.

sql/parse/parse.go
ajnavarro: should not, but can be reached, and an error is better than a panic if possible, in mi opinion.

erizocosmico: Can't because there are no other types implemented for SQLVal unless you set a value that's not correct on a node (which won't happen because they're returned by vitess and we control vitess versioning). But I will change it to an error, anyway.

mcarmonaa:
What is the state of this? Is it expected to be merged or it will be merged in a next version with the changes commented? :thinking:

erizocosmico:
As soon as we have all the basic features for the MVP I can give this some love and keep working on it. But yeah, not expected to be merged right away.

erizocosmico:
Fixed @mcarmonaa @ajnavarro

erizocosmico:
Also implemented IN and NOT IN.

ajnavarro:
After the small pending change, LGTM.

erizocosmico:
Done @ajnavarro

PR #72, opened 2018-02-28
sql/(parse,plan): implement inner join

### LINKED ISSUE #57: TPC-H: Implement Inner Join

erizocosmico:
Closes #57

Implements the INNER JOIN node, which right now is basically a `Filter(cond, CrossJoin(left, right))` given it has no indexes whatsoever.

The only clause of the INNER JOIN implemented is the ON clause. USING does not make sense to implement until we have a proper handling of column qualifiers.

The implementation of INNER JOIN differs from the one in CROSS JOIN. While the cross join loads all the right rows in memory and iterates the left rows, the inner join creates a new iterator for each iteration of the left iterator, saving huge amounts of memory on large tables.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -77,6 +77,15 @@ func TestQueries(t *testing.T) {
 		"SELECT i FROM mytable WHERE i NOT BETWEEN 1 AND 2",
 		[][]interface{}{{int64(3)}},
 	)
+
+	testQuery(t, e,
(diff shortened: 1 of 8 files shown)


erizocosmico:
**Update:**
Cross join iterator no longer keeps all right rows in memory. Instead it uses the implementation previously used by inner join.
Inner join now makes use of the cross join iterator and the filter iterator.

ajnavarro:
Rebase please

erizocosmico:
Rebased

PR #81, opened 2018-03-06
[WIP] integration test with queries from QuerySetApp

mcarmonaa:
This test takes those queries from `QuerySetApp` that don't use UDFs. Apart of the expressions already implemented they need `CASE` and `HAVING` so we should implement them if you don't know other way to formulate the queries.

If everything is ok with the queries, I'll add the logic of the database creation and prepare a known data against to test the queries.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -1,0 +1,39 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	sqle "gopkg.in/src-d/go-mysql-server.v0"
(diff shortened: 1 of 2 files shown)


erizocosmico:
Shouldn't this go in gitquery? You're mentioning repositories and history_idx and so on and this is the generic implementation of a mysql server, it does not have anything to do with those things, that will be implemented in gitquery.

ajnavarro:
yes, this should be on gitquery

mcarmonaa:
Oops! right, moving there

PR #84, opened 2018-03-07
Add CREATE TABLE DDL statement

ajnavarro:
Also, memory DB implementation constructor now returns the sql.Database interface instead of the implementation itself.

Signed-off-by: Antonio Jesus Navarro Perez <antnavper@gmail.com>

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -104,17 +104,26 @@ func qualifyColumns(a *Analyzer, n sql.Node) (sql.Node, error) {
 }
 
 func resolveDatabase(a *Analyzer, n sql.Node) (sql.Node, error) {
-	_, ok := n.(*plan.ShowTables)
-	if !ok {
-		return n, nil
+	// TODO Database should implement node,
+	// and ShowTables and CreateTable nodes should be binaryNodes
+	switch v := n.(type) {
+	case *plan.ShowTables:
+		db, err := a.Catalog.Database(a.CurrentDatabase)
+		if err != nil {
+			return n, err
+		}
+
+		v.Database = db
+	case *plan.CreateTable:
+		db, err := a.Catalog.Database(a.CurrentDatabase)
+		if err != nil {
+			return n, err
+		}
+
+		v.Database = db
 	}
 
-	db, err := a.Catalog.Database(a.CurrentDatabase)
-	if err != nil {
-		return n, nil
-	}
-
-	return plan.NewShowTables(db), nil
+	return n, nil
 }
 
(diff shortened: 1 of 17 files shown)


ajnavarro:
Missing some tests.

sql/analyzer/rules.go
```diff
@@ -104,17 +104,26 @@ func qualifyColumns(a *Analyzer, n sql.Node) (sql.Node, error) {
...
 
 func resolveDatabase(a *Analyzer, n sql.Node) (sql.Node, error) {
-	_, ok := n.(*plan.ShowTables)
-	if !ok {
-		return n, nil
-	}
+	// TODO Database should implement node,
+	// and ShowTables and CreateTable nodes should be binaryNodes
+	switch v := n.(type) {
+	case *plan.ShowTables:
```
jfontan: Can we use `fallthrough` so we don't repeat the same code in both cases?

sql/plan/ddl.go
```diff
@@ -0,0 +1,59 @@
...
+	name     string
+	schema   sql.Schema
+}
+
+// NewCreateTable creates a new CreateTable node
+func NewCreateTable(db sql.Database, name string, schema sql.Schema) *CreateTable {
+	return &CreateTable{
+		Database: db,
+		name:     name,
+		schema:   schema,
```
jfontan: I'm not sure it's needed but here we can set `Source` for each column with the name of the table.

erizocosmico:
Yep, is needed
On Mon, 12 Mar 2018 at 17:19, Javi Fontan <notifications@github.com<mailto:notifications@github.com>> wrote:

@jfontan commented on this pull request.

________________________________

In sql/plan/ddl.go<https://github.com/src-d/go-mysql-server/pull/84#discussion_r173857092>:

> +     "gopkg.in/src-d/go-mysql-server.v0/sql<http://gopkg.in/src-d/go-mysql-server.v0/sql>"
+)
+
+// CreateTable is a node describing the creation of some table.
+type CreateTable struct {
+       Database sql.Database
+       name     string
+       schema   sql.Schema
+}
+
+// NewCreateTable creates a new CreateTable node
+func NewCreateTable(db sql.Database, name string, schema sql.Schema) *CreateTable {
+       return &CreateTable{
+               Database: db,
+               name:     name,
+               schema:   schema,


I'm not sure it's needed but here we can set Source for each column with the name of the table.

sql/analyzer/rules.go
ajnavarro: Is not possible to fallthrough in a type switch

PR #83, opened 2018-03-06
sql/analyzer: implement projection pushdown

erizocosmico:
Depends on #82 

Closes #27

PR #86, opened 2018-03-07
sql/(plan,analyzer): implement pushdown of filters and cols

### LINKED ISSUE #27: Pushdown Projections to Tables
To do not return all the columns of a table if it is not necessary, we should push down the actual needed projection to tables iterators to filter before hand.

### LINKED ISSUE #28: Pushdown filters to Tables
To filter before hand rows to improve performance, we should push down filters to Tables.

erizocosmico:
Depends on #82 

Closes #27 
Closes #28 

Supersedes #83 

### Implementation details

Only columns and filters that apply to a column are pushed down to that column. That is, if you have `42 > 5` that filter will not be pushed down to any table, and if you have `a.foo = 5` that filter will only be pushed down to table `a`.

If more than one table is mentioned in a filter, it's not pushed down to neither. E.g. `a.foo = b.foo` is not pushed down.

Perhaps we want the opposite and we want all filters pushed down to the table, regardless of whether it mentions the table or not?

Also, join conditions are not being pushed down as filters, should we also push those down?

### DIFF
--- a/sql/analyzer/filters.go
+++ b/sql/analyzer/filters.go
@@ -1,0 +1,84 @@
+package analyzer
+
+import (
+	"reflect"
+
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+)
+
+type filters map[string][]sql.Expression
+
+func (f filters) merge(f2 filters) {
+	for k, exprs := range f2 {
+		f[k] = append(f[k], exprs...)
+	}
+}
+
+func exprToTableFilters(expr sql.Expression) filters {
+	filtersByTable := make(filters)
+	for _, expr := range splitExpression(expr) {
+		var tables []string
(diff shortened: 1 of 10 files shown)


sql/analyzer/filters.go
```diff
@@ -0,0 +1,84 @@
...
+			f, ok := e.(*expression.GetField)
+			if ok {
+				tables = append(tables, f.Table())
+			}
+
+			return e, nil
+		})
+
+		if len(tables) == 1 {
+			filtersByTable[tables[0]] = append(filtersByTable[tables[0]], expr)
```
jfontan: I believe I don't understand this logic. Does it mean that the expression is only added to the left table?

erizocosmico: A filter is only pushed down if only one table is mentioned in the expression (unless we want to change that behavior)

jfontan: :+1:

erizocosmico:
Updated to receive sessions

PR #87, opened 2018-03-07
server: implement session manager

erizocosmico:
With the work done in #80 we can pass sessions around, but there is no way to tell the server to create custom sessions. This implements a Session Manager that keeps track of all the sessions and handles cancellation of sessions when connection is closed as well as a mechanism to create custom sessions using SessionBuilders.


erizocosmico:
Rebased against master

PR #89, opened 2018-03-09
 sql/analyzer: resolve GetField on schemas with repeated column names 

### LINKED ISSUE #88: ambiguous columns cannot be used
Imagine we have table `foo` with column `a` and table `bar` with column `a` as well. Right now, it's not possible to use both columns.

**NOTE:** you can `select *` a result with ambiguous columns, you just cannot use them e.g. `select foo.a, foo.b`.

**NOTE 2:** PR to fix this should start from #82

erizocosmico:
Depends on #82 

Fixes #88

Because of the way we access the fields in the rows using GetField with the index in which they appear and only by name, there was no way of knowing which field was the correct when there were multiple fields with the same name.

Now, all columns include a "Source", which is defined by the table and must match the table name (this is validated by an analyzer validation rule), so that we can identify from which table we should get the field.

One notable exception to this is TableAlias, which avoids this limitation only when its child node is not a Projection (that is, a subquery).

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -132,6 +132,58 @@ func TestInsertInto(t *testing.T) {
 	)
 }
 
+func TestAmbiguousColumnResolution(t *testing.T) {
+	require := require.New(t)
+
+	table := mem.NewTable("foo", sql.Schema{
+		{Name: "a", Type: sql.Int64, Source: "foo"},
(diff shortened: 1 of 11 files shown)


ajnavarro:
Rebase before merge please

PR #90, opened 2018-03-09
sql/(plan,analyzer): implement OrderedDistinct for sorted results

erizocosmico:
This improves the performance of the Distinct operation by two orders of magnitude (both in CPU and memory) when the result set is sorted.

Also adds an `Equals` method to `Row`, which can come in handy for some other operations.

Benchmark results:
```
BenchmarkDistinct-4          	     100	  13723600 ns/op	 2183356 B/op	  264693 allocs/op
BenchmarkOrderedDistinct-4   	   10000	    117517 ns/op	   65904 B/op	     925 allocs/op
BenchmarkProject-4           	   20000	     99851 ns/op	   65324 B/op	     913 allocs/op
PASS
ok  	gopkg.in/src-d/go-mysql-server.v0/sql/plan	5.600s
```

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -17,6 +17,7 @@ var DefaultRules = []Rule{
 	{"resolve_database", resolveDatabase},
 	{"resolve_star", resolveStar},
 	{"resolve_functions", resolveFunctions},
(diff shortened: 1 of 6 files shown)


erizocosmico:
Rebased

PR #94, opened 2018-03-12
[WIP] Add logictests from SQLite

ajnavarro:
Depends on #143

- Implement the parser as the standard
- Added a flag (FIXME) to skip queries that are failing
- All the tests that throws `unsupported syntax` error, are ignored


ajnavarro:
Depends on #143

PR #100, opened 2018-03-13
sql/analyzer: use comparison operator instead of reflect.DeepEqual

erizocosmico:
Due to the immutable nature of the execution plan, any change in any part of the tree means the outermost node will change, so we don't need to use `!reflect.DeepEqual` but just the `!=` operator for doing the comparison in the analyzer.

As long as the pointer to the node is the same, the two `Node` interface values are equal.

Proof: https://play.golang.org/p/Jb2rya-enQO

### DIFF
--- a/sql/analyzer/analyzer.go
+++ b/sql/analyzer/analyzer.go
@@ -2,7 +2,6 @@ package analyzer
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


erizocosmico:
Ok, nvm, it fails big time.

PR #103, opened 2018-03-14
Implement CONVERT expression and cast expressions in comparisons.

### LINKED ISSUE #92: Panic comparing literals and columns numbers
If we have the following query:
`SELECT a from b where a > 10`
It will panic because `a` is an int32 and the literal `10` is an int64 by default. Line on parse.go:
```go
return expression.NewLiteral(n, sql.Int64), nil
```

The standard solution for this is add a `CAST` node surrounding the literal, casting it to the type of the other expression into the comparison. Doing this, all the those queries will be executable:
`SELECT a from b where a > 10`
`SELECT a from b where a > "10"`
`SELECT a from b where a > 10.0`
`SELECT a from b where 10 < a`

mcarmonaa:
Closes #92


sql/analyzer/rules.go
```diff
@@ -411,3 +412,71 @@ func pushdown(a *Analyzer, n sql.Node) (sql.Node, error) {
 		return node, nil
 	})
 }
+
+func castCompares(a *Analyzer, n sql.Node) (sql.Node, error) {
```
erizocosmico: What do you think about putting this logic inside the `Comparison` `Compare` method instead?

`Compare` has access to the types and it could handle them itself using the `Convert` method of the types. The less rules we need, the better.

sql/expression/comparison.go
```diff
@@ -236,25 +259,25 @@ func (lt *LessThan) TransformUp(f func(sql.Expression) (sql.Expression, error))
...
+	return &GreaterThanOrEqual{newComparison(left, right)}
 }
 
 // Eval implements the Expression interface.
 func (gte GreaterThanOrEqual) Eval(
 	session sql.Session,
 	row sql.Row,
 ) (interface{}, error) {
-	a, err := gte.Left.Eval(session, row)
+	a, err := gte.BinaryExpression.Left.Eval(session, row)
```
erizocosmico: why add the `BinaryExpression` here?

sql/expression/comparison.go
```diff
@@ -6,51 +6,74 @@ import (
...
 
-// Comparison is an expression that compares an expression against another.
-type Comparison struct {
+// Compararer implements a comparison expression.
+type Comparer interface {
+	sql.Expression
+	IsComparison() bool
+	Left() sql.Expression
+	Right() sql.Expression
+	SetLeft(sql.Expression)
```
erizocosmico: nodes should be immutable. If it needs to change, a new node should be created

sql/expression/convert.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Name implements the Expression interface.
+func (c *Convert) Name() string {
+	return fmt.Sprintf("convert(%v, %v)", c.Child.Name(), c.castToType)
+}
+
+// TransformUp implements the Expression interface.
+func (c *Convert) TransformUp(f func(sql.Expression) (sql.Expression, error)) (sql.Expression, error) {
+	child, err := c.UnaryExpression.Child.TransformUp(f)
```
erizocosmico: `UnaryExpression`  here is redundant

sql/expression/convert.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return f(NewConvert(child, c.castToType))
+}
+
+// Eval implements the Expression interface.
+func (c *Convert) Eval(session sql.Session, row sql.Row) (interface{}, error) {
+	val, err := c.UnaryExpression.Child.Eval(session, row)
```
erizocosmico: `UnaryExpression` here is redundant

sql/expression/comparison.go
mcarmonaa: Because of the `Left()` method in the interface, it shadows the `Left` field of the embedded `BinaryExpression`.

sql/analyzer/rules.go
ajnavarro: The rule that I'm missing here is one to cast literals in comparisons when the types on left and right are not equals.

sql/analyzer/rules.go
```diff
@@ -411,3 +412,71 @@ func pushdown(a *Analyzer, n sql.Node) (sql.Node, error) {
...
+		}
+
+		left, right = convertExpressionsToText(left, right)
+		c.SetLeft(left)
+		c.SetRight(right)
+		return c.(sql.Expression), nil
+	})
+}
+
+func isNumber(t sql.Type) bool {
```
ajnavarro: Maybe this logic to check types should be in types.go file

sql/analyzer/rules_test.go
```diff
@@ -396,6 +396,128 @@ func (t *pushdownProjectionAndFiltersTable) TransformUp(f func(sql.Node) (sql.No
...
+						expression.NewGetField(1, sql.Int32, "number", false),
+						expression.NewLiteral(int64(2), sql.Int64),
+					),
+				},
+				plan.NewUnresolvedTable("foo"),
+			),
+			expected: plan.NewProject(
+				[]sql.Expression{
+					expression.NewGreaterThanOrEqual(
+						expression.NewConvert(
```
ajnavarro: The auto casting should be always only on the literal side I think

sql/expression/comparison.go
```diff
@@ -6,51 +6,74 @@ import (
...
+	SetRight(sql.Expression)
+}
+
+type comparison struct {
```
ajnavarro: this should be public, to be able to check if the expression is a Comparison on rules.

engine_test.go
```diff
@@ -127,6 +127,65 @@ func TestQueries(t *testing.T) {
...
+	testQuery(t, e,
+		"SELECT '3' > 2 FROM mytable",
+		[][]interface{}{
+			{true},
+			{true},
+			{true},
+		},
+	)
+
+	testQuery(t, e,
```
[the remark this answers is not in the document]
mcarmonaa: yep

sql/analyzer/rules.go
mcarmonaa: `Literal`s are like any other `Expression` and they have certain type. The rule can't be cast always the literal to the other type, at least mysql doesn't work in this way.

For example, when the types at both side of the comparison are different, mysql cast to numbers if some of the type of them is a number, independently if there is a literal or not.

As you can see in this example, mysql cast the text values from the `num` column to numbers, (the casting of the text values evals to 0 because they don't represent numbers):
```
MySQL [testing]> select * from tabletest;
+----+------+------+
| id | name | num  |
+----+------+------+
|  1 | a    |    1 |
|  2 | b    |    2 |
|  3 | c    |    3 |
|  4 | d    |    4 |
+----+------+------+
4 rows in set (0.00 sec)

MySQL [testing]> select * from tabletest where name > 0;
Empty set (0.00 sec)

MySQL [testing]> select * from tabletest where name < 0;
Empty set (0.01 sec)

MySQL [testing]> select * from tabletest where name = 0;
+----+------+------+
| id | name | num  |
+----+------+------+
|  1 | a    |    1 |
|  2 | b    |    2 |
|  3 | c    |    3 |
|  4 | d    |    4 |
+----+------+------+
4 rows in set (0.00 sec)

```

sql/analyzer/rules_test.go
mcarmonaa: As explained up there in the other comment, the cast can't be just for literals and should be for both expressions because of the "types" the convert expression accepts.

Anyway, I'm moving all the logic to Comparison expression and removing the rule as @erizocosmico suggested.

sql/expression/comparison.go
mcarmonaa: That check is done by the interface `Comparer`, because `comparison` is just used as an embedded type in the real comparison expressions.

If you try a type assertion on a interface by the embedded type of the type, it doesn't work.
See: https://play.golang.org/p/NJHOB7G27mc

mcarmonaa:
@erizocosmico @ajnavarro I removed the rule and moved the casting types logic to the comparison expressions

sql/expression/comparison.go
```diff
@@ -3,73 +3,165 @@ package expression
 import (
 	"regexp"
 
+	errors "gopkg.in/src-d/go-errors.v0"
```
erizocosmico: we should be using errors v1

sql/expression/comparison.go
```diff
@@ -3,73 +3,165 @@ package expression
...
-type Comparison struct {
+// Compararer implements a comparison expression.
+type Comparer interface {
+	sql.Expression
+	Compare(session sql.Session, row sql.Row) (int, error)
+	Left() sql.Expression
+	Right() sql.Expression
+}
+
+var ErrNilOperand = errors.NewKind("nil operand found in comparison")
```
erizocosmico: comment missing for this

sql/expression/convert.go
```diff
@@ -0,0 +1,212 @@
...
+	casted, err := ConvertValue(val, c.castToType)
+	if err != nil {
+		return nil, ErrConvertExpression.Wrap(err, c.Name(), c.castToType)
+	}
+
+	return casted, nil
+}
+
+// ConvetValue cast the value of val to castTo throw the Convert expression rules.
+func ConvertValue(val interface{}, castTo string) (interface{}, error) {
```
erizocosmico: is this used outside this package? If it's not, I'd keep it lowercase

sql/expression/convert.go
```diff
@@ -0,0 +1,212 @@
...
+	errors "gopkg.in/src-d/go-errors.v0"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// ErrConvertExpression is returned when a conversion is not possible.
+var ErrConvertExpression = errors.NewKind("expression '%v': couldn't convert to %v")
+
+const (
+	// ConvertTo represents those conversion types Convert accepts.
+	ConvertToBinary   = "binary"
```
erizocosmico: are these constants used outside this package? If not, I'd keep them lowercase. And if so, I would make them integers, the take less space and the string is not really needed. They're only needed as an enum

sql/expression/comparison.go
mcarmonaa: Should I change it across all the repo?

ajnavarro: Change only the added ones, I opened an issue for the rest: https://github.com/src-d/go-mysql-server/issues/115

sql/expression/convert.go
mcarmonaa: Those are the string go-vitess' parser returns, I can change it to enums, but I'd have to create two maps kept in memory to translate strings->int and int->strings and make the logic a little bit complex.

They are only used inside the expression package, but I think it's ok to have them public if at some point we need to create `Convert` expressions in a rule or something like that. 

Do you think it will make a really improve the enum thing?

erizocosmico: Then I think the enum is not worth it, too much hassle for so little benefit. LGTM as is, then.

PR #110, opened 2018-03-15
Change assertion to check content without order

ajnavarro:
Is strange, but in some cases this tests fails because the query output is in a different order than expected.
That is ok, if the query does not have a specific order, the order is not ensured.


mcarmonaa:
We should merge this asap to avoid others PRs failing

erizocosmico:
Gonna merge this, otherwise we will keep having random builds failing

PR #111, opened 2018-03-15
sql/(plan,expression): implement String method to make tree printable

### LINKED ISSUE #95: Create String() method on Node Interface
Create a `String()` method on `Node` interface to be able to print an entire tree. This can be really useful when something fails and we throw an error, or for debugging.

`Expression` has a `Name()` method that should be changed to `String()` too.

### LINKED ISSUE #108: Plan pretty printer
We have a #95 to add `String()` method to all nodes so they can be printed, but that will not generate a tree-like structure to pretty-print the plan as such.

We might want to have a way to pretty-print the plan in a tree structure.

When we have the `Children` method in expressions, as described in #107, and the `String` function in all nodes we could do a generic function to pretty-print the plan as a tree.

Something like

```
Project(foo)
|- Filter(yada yada)
    |- CrossJoin
        |- Table(bar)
        |- Table(baz)
```

API could by like:

```go
result := plan.Pretty(node)
```

erizocosmico:
Closes #95
Closes #108

- Implements `String` method on all `sql.Node`s.
- Implements `String` method on all `sql.Expression`s.
- Remove `Name` method from `sql.Expression`s.
- Only leave `Name` in expressions that will be `sql.Nameable`.
- Created `TreePrinter` to print trees with a node and children.


engine_test.go
```diff
@@ -293,3 +294,27 @@ func newEngine(t *testing.T) *sqle.Engine {
...
+ └─ Limit(5)
+     └─ Project(t.foo, bar.baz)
+         └─ Filter(foo > qux)
+             └─ InnerJoin(foo = baz)
+                 ├─ TableAlias(t)
+                 │   └─ UnresolvedTable(tbl)
+                 └─ UnresolvedTable(bar)
+`
+
+func TestPrintTree(t *testing.T) {
```
ajnavarro: Should be this test on parse package?

mem/table.go
```diff
@@ -72,3 +72,7 @@ func (t *Table) Insert(row sql.Row) error {
 	t.data = append(t.data, row.Copy())
 	return nil
 }
+
+func (t Table) String() string {
+	return fmt.Sprintf("Table(%s)", t.name)
```
ajnavarro: Maybe should we print the schema here too?

sql/plan/cross_join.go
```diff
@@ -75,6 +75,13 @@ func (p *CrossJoin) TransformExpressionsUp(f func(sql.Expression) (sql.Expressio
 	return NewCrossJoin(left, right), nil
 }
 
+func (p *CrossJoin) String() string {
+	pr := sql.NewTreePrinter()
+	_ = pr.WriteNode("CrossJoin")
```
ajnavarro: Should we print the conditions too?

engine_test.go
erizocosmico: I'm only using parse to not create the tree by hand. Since it's kind of an integration test checking some nodes return the correct output, I thought of putting it on the engine test

mem/table.go
erizocosmico: Sure

sql/plan/cross_join.go
erizocosmico: CrossJoin does not have conditions (at least ours doesn't)

ajnavarro: oops, sorry.

sql/plan/pushdown.go
```diff
@@ -77,3 +89,26 @@ func (t *PushdownProjectionAndFiltersTable) TransformUp(f func(sql.Node) (sql.No
...
+		columns[i] = col.String()
+	}
+
+	var filters = make([]string, len(t.filters))
+	for i, f := range t.filters {
+		filters[i] = f.String()
+	}
+
+	_ = pr.WriteChildren(
+		fmt.Sprintf("Columns(%s)", strings.Join(columns, ", ")),
```
ajnavarro: PushedDownColumns to make it clear?

erizocosmico:
Don't worry, the fixes in gitquery are easy and I will have them right away.

PR #125, opened 2018-03-20
Walk the tree

### LINKED ISSUE #106: Add Walk function to plan
# Problem

We have many rules in the analyzer that need to gather some data from the whole tree before making any transformations. To do that, right now they make use of `TransformUp`, which creates a lot of nodes that are unnecessary, since we are not transforming anything, just gathering data.

# Solution

It would be very handy to have a `Walk` function in the plan that allowed us to traverse the plan nodes.

Why a function instead of a method? To not add more and more methods to the `Node` interface. Besides, we don't need all nodes to implement that, only the ones we care about. For example, we don't need table implementations to provide their `Walk` method. Plan nodes can only be generated by us, since we are the ones parsing.

Besides, we have the `Children` method in the `Node` interface, which makes having a per-node implementation of `Walk` irrelevant.

Ideal API would be the following:

```go
plan.Walk(node, func(sql.Node) plan.Visitor {
        // do something with the node
})
```

The callback should return a `Visitor` with perhaps custom logic for the given node or nil to stop.

Then the shortcut:

```go
plan.Inspect(node, func(sql.Node) bool {
        // do something with the node
})
```
The callback should return a boolean. If the result is `true` it will stop walking.

It should work like [go/ast#Inspect](https://golang.org/pkg/go/ast/#Inspect) and [go/ast#Walk](https://golang.org/pkg/go/ast/#Walk)

### LINKED ISSUE #107: Add Walk function to expression
Same as https://github.com/src-d/go-mysql-server/issues/106, but for expressions.

The only thing that I would change here is make `Expression` have a `Children` method as `Node` does. With this, we can have a function that works generically on all nodes without having to force all `Expression` implementations to implement their `Walk` method.

erizocosmico:
Closes #106 
Closes #107 

It only diverges from the original proposal in that it adds `WalkExpressions` and `InspectExpressions` in `plan` package because of the limitation of `Expression`s not being `Node`s.

### DIFF
--- a/sql/plan/walk.go
+++ b/sql/plan/walk.go
@@ -52,40 +52,9 @@ func Inspect(node sql.Node, f func(sql.Node) bool) {
 // expression it finds.
 func WalkExpressions(v expression.Visitor, node sql.Node) {
 	Inspect(node, func(node sql.Node) bool {
-		switch node := node.(type) {
-		case *Project:
-			for _, e := range node.Expressions {
+		if node, ok := node.(sql.Expressioner); ok {
+			for _, e := range node.Expressions() {
 				expression.Walk(v, e)
-			}
-		case *Filter:
-			expression.Walk(v, node.Expression)
-		case *Values:
-			for _, tuple := range node.ExpressionTuples {
-				for _, e := range tuple {
-					expression.Walk(v, e)
-				}
-			}
-		case *PushdownProjectionAndFiltersTable:
-			for _, f := range node.columns {
-				expression.Walk(v, f)
-			}
-
-			for _, f := range node.filters {
-				expression.Walk(v, f)
-			}
-		case *GroupBy:
-			for _, e := range node.Aggregate {
-				expression.Walk(v, e)
-			}
-
-			for _, e := range node.Grouping {
-				expression.Walk(v, e)
-			}
-		case *InnerJoin:
-			expression.Walk(v, node.Cond)
-		case *Sort:
-			for _, f := range node.SortFields {
-				expression.Walk(v, f.Column)
 			}
 		}
 		return true
--- a/sql/plan/walk.go
+++ b/sql/plan/walk.go
@@ -1,0 +1,46 @@
+package plan
+
+import "gopkg.in/src-d/go-mysql-server.v0/sql"
+
+// Visitor visits nodes in the plan.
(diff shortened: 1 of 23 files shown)


sql/plan/walk.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+)
+
+// Visitor visits nodes in the plan.
+type Visitor interface {
+	// Visit method is invoked for each node encountered by Walk.
+	// If the result Visitor is not nul, Walk visits each of the children
```
ajnavarro: s/nul/nil

sql/plan/walk.go
```diff
@@ -0,0 +1,108 @@
...
+// f(node); node must not be nil. If f returns true, Inspect invokes f
+// recursively for each of the children of node, followed by a call of
+// f(nil).
+func Inspect(node sql.Node, f func(sql.Node) bool) {
+	Walk(inspector(f), node)
+}
+
+// WalkExpressions traverses the plan and calls expression.Walk on any
+// expression it finds.
+func WalkExpressions(v expression.Visitor, node sql.Node) {
```
ajnavarro: The content of this function is a little bit specific, right? Maybe we should add a method on Expression interface to return all the expressions on that Expression implementation.

erizocosmico: `Expression` already has the `Children` method to do that. The problem is that you can't access the expressions of a `Node` because you'd need to do a lot of work every single time you want to do that (just like this function does).

ajnavarro: Then we need something to avoid have new expressions in the future that are not implemented here, causing a bad walking result. What do you think?

erizocosmico: We could have an `Expressioner` interface with the method `Expressions` that does the same as `Children` but returning expressions. Instead of forcing it on the `Node` interface I would make it optional, so only the types that have expressions need to implement it.

ajnavarro: Sounds good, any thoughts @jfontan @mcarmonaa ?

erizocosmico:
Fixed @ajnavarro

PR #126, opened 2018-03-20
Fix unexpected error when an empty set is returned

### LINKED ISSUE #122: Unexpected query ended
If the query has no results, the next error is thrown by the mysql client:

`ERROR 1105 (HY000): unexpected: query ended without no results and no error`

mcarmonaa:
Closes #122 

Signed-off-by: Manuel Carmona <manu.carmona90@gmail.com>

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -92,6 +92,7 @@ func (h *Handler) ComQuery(
 	}
 
 	var r *sqltypes.Result
+	var proccesedAtLeastOneBatch bool
 	for {
 		if r == nil {
 			r = &sqltypes.Result{Fields: schemaToFields(schema)}
@@ -103,6 +104,7 @@ func (h *Handler) ComQuery(
 			}
 
 			r = nil
+			proccesedAtLeastOneBatch = true
 
 			continue
 		}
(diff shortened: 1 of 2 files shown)


server/handler_test.go
```diff
@@ -42,12 +42,14 @@ func TestHandlerOutput(t *testing.T) {
...
-		lastRows = len(res.Rows)
+		if res.RowsAffected != 0 {
+			lastRowsAffected = res.RowsAffected
+			lastRows = len(res.Rows)
+		}
 		return nil
 	})
 	require.NoError(err)
-	require.Equal(1, c)
+	require.Equal(2, c)
```
ajnavarro: In this case, with a row batch of 100, it should be called only once. In this case, the second call is useless.

server/handler.go
```diff
@@ -120,7 +122,11 @@ func (h *Handler) ComQuery(
 		r.RowsAffected++
 	}
 
-	if r != nil && r.RowsAffected != 0 {
+	// Even if r.RowsAffected = 0, the callback must be
+	// called to update the state in the go-vitess' listener
+	// an avoid to return errors when the query doesn't
```
erizocosmico: s/an/and/
s/avoid to return/avoid returning/

server/handler.go
```diff
@@ -120,7 +122,11 @@ func (h *Handler) ComQuery(
...
+	// produce results.
```
erizocosmico: s/results/any results/

mcarmonaa:
@ajnavarro ping

PR #128, opened 2018-03-20
sql/expression: split functions and aggregations

### LINKED ISSUE #67: Split UDFs into several files in their own package
Right now, we have all functions inside `sql/expression` in a file called `function.go`. Since functions are not just an isolated function but a struct with many methods, this can go out of hand when we start adding a lot of functions even though right now seems manageable.

I propose adding a new package `sql/expression/function` and put every function inside a different file. That way its easier to find a specific function and they're not lying around a file all together.

Same with aggregations when we have enough of them that having them in a file starts to be unmanageable.

erizocosmico:
Closes #67


sql/expression/function/registry.go
```diff
@@ -0,0 +1,39 @@
...
+	"month":     sql.Function1(NewMonth),
+	"day":       sql.Function1(NewDay),
+	"hour":      sql.Function1(NewHour),
+	"minute":    sql.Function1(NewMinute),
+	"second":    sql.Function1(NewSecond),
+	"dayofyear": sql.Function1(NewDayOfYear),
+}
+
+// RegisterDefaults registers the aggregations in the catalog.
+func RegisterDefaults(c *sql.Catalog) error {
```
mcarmonaa: Maybe we should export the list of functions and make the `Catalog` has a method to register a list of functions.

erizocosmico: Yeah, we could do that. Anyone against that change? @jfontan @ajnavarro

ajnavarro: @erizocosmico sounds good to me too.

erizocosmico:
Done @ajnavarro @mcarmonaa

sql/functionregistry.go
```diff
@@ -122,6 +122,9 @@ func (FunctionN) isFunction() {}
 // and User-Defined Functions.
 type FunctionRegistry map[string]Function
```
jfontan: Should this be changed to `type FunctionRegistry Functions`?

erizocosmico: Not really. They just happen to point to the same underlying type, but they have different purposes. `Functions` is just a shortcut and `FunctionRegistry` is the one with the methods and so on. I mean, it _could_ be changed, but it doesn't bring any gain, except that it may change in the future while `Functions` will probably not.

PR #129, opened 2018-03-20
*: implement a configuration mechanism

### LINKED ISSUE #109: Add configuration per engine instance and per session
We need a way to set configurations at the engine (or BBDD) and session levels. This configuration should be of two types, one specific to configure parameters on go-mysql-server (log level, number of rows to return per batch on a connection) or specific depending on the database implementation (on the gitBase case, the max size of files to return in blob table)

erizocosmico:
Closes #109 

This only implements the Config API and introduces it into the `Session` and `Engine`, we still need to figure out how to fill that configuration (files, environment variables, `set` commands, ...).

### Beware!

This introduces breaking changes that will make gitquery explode, so don't merge without caution.

### Config API

This API is designed for config to be inheritable, type-safe and thread-safe. For our use case, what we need is the following:
- Be able to work in multiple goroutines (when the engine works in parallel)
- Be able to have an Engine configuration and a Session configuration where the Session configuration can access the configuration values from the Engine and override them without actually changing them. That's why the `Config` has a parent that allows it to access it, while not being able to change it.
- Not panic but be able to store any kind of value. Hence, the `Set*` and `*` methods for getting the different types of values and the error in the getters.

To avoid the "what happens when there is no config? do we throw an error?" all methods require a default value when you're getting the config value.

The only part I don't feel much of this API is the usage from the `Session`. From a session it would be used like this:

```go
val, err := session.Config().String("my_key", "default")
```

One option we could explore to overcome this is making `Config` an interface and embed it on the `Session` interface.

The previous snippet would become:

```go
val, err := session.String("my_key", "default")
```

Which, to be honest, is actually a bit weird.

Another option we could explore is just duplicate methods on session:

```go
val, err := session.StringConfig("my_key", "default")
session.SetStringConfig("my_key", "my_val")
```

### Next steps

- Figure out how to let the user fill this config from the outside

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,144 @@
+package config
+
+import (
+	"sync"
+
+	errors "gopkg.in/src-d/go-errors.v0"
+)
+
+// Config is a thead-safe container for the application's configuration.
+// The configuration can have a parent configuration to be able to inherit
+// config values from it.
+type Config struct {
+	mu     sync.RWMutex
+	parent *Config
+	kv     map[string]interface{}
+}
+
+// New creates an empty configuration.
+func New() *Config {
+	return &Config{kv: make(map[string]interface{})}
+}
+
+// FromConfig creates a new configuration from the given parent. This parent
+// will be used only to read and will never be modified.
+func FromConfig(c *Config) *Config {
+	return &Config{parent: c, kv: make(map[string]interface{})}
+}
+
+func (c *Config) set(k string, v interface{}) {
+	c.mu.Lock()
+	c.kv[k] = v
+	c.mu.Unlock()
+}
+
+func (c *Config) get(k string) (interface{}, bool) {
+	c.mu.RLock()
+	v, ok := c.kv[k]
+	c.mu.RUnlock()
+
+	if !ok && c.parent != nil {
+		return c.parent.get(k)
+	}
+
+	return v, ok
+}
+
+// SetString stores a string value at key k.
+func (c *Config) SetString(k string, v string) {
+	c.set(k, v)
+}
+
+// SetInt stores an int64 value at key k.
+func (c *Config) SetInt(k string, v int64) {
+	c.set(k, v)
+}
+
+// SetBool stores a bool value at key k.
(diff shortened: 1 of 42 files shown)


config/config.go
```diff
@@ -0,0 +1,144 @@
+package config
+
+import (
+	"sync"
+
+	errors "gopkg.in/src-d/go-errors.v0"
```
mcarmonaa: Use go-errors.v1 instead

erizocosmico:
Fixed, @mcarmonaa

erizocosmico:
Right now it's not set, I was waiting until we knew what kind of configs we wanted to support.

What do you think of the following:

- Add the following methods: `LoadFromFile(path string) error` and `LoadFromEnv(vars ...string) error` (note the `vars ...string`, without that, we would be loading all the os env into the config when it's not needed). With this, an implementor can decide how and what things to load into config. Perhaps we could still load some initial things that are always needed in go-mysql-server and so on.
- Add support for `SET k=v` queries and write that into the config.

So, I guess we want to have a base config that comes from go-mysql-server (which avoids every implementor to load certain things) and then a config given by the implementor to the `Handler`.
Since `Config` accepts parents, it would be easy to make the base config the parent of the implementor config, which will be the parent of the connection config, that will be the parent of the context config.

Another important thing: at what level should `SET` work? context? connection? global? The config that will be passed with the context is for context level, so it will not be visible by other contexts in the connection. If that's not our use case we need a mechanism to be able to write to the parents (probably just make `Parent` public) for `SET` commands.

Thoughts? /cc @mcarmonaa @jfontan @ajnavarro @smola

jfontan:
There are global and session variables in mysql:

https://dev.mysql.com/doc/refman/5.7/en/set-variable.html

For me it makes sense to have global and connection variables. If I set a variable I expect to have it set while I don't disconnect.

erizocosmico:
Implemented `SET` command for setting config values via queries.

erizocosmico:
Closing, since it's not needed.

PR #133, opened 2018-03-21
*: make Session per connection and Context per query

### LINKED ISSUE #132: Session souldn't be created per query
Right now, we are creating a new session on every executed query. 

In my opinion, we should have a session per connection. That allows us to have some configuration attached to the actual session of the user, that can be applied to all the queries that are being executed on that connection.

One problem with that is right now a Session is also a Context. If we cancel the context, that session cannot be used again.

My proposal is to move the context outside the Session interface, in a way that you can set one on every query:

```go
type Session interface {
 SetContext(context.Context)
 GetContext()context.Context
}
```

Other solution could be separate completely the session and the context, sending both to the Nodes:

```go
type Expression interface {
[...]
	Eval(Session, Context, Row) (interface{}, error)
[...]
}
```

What do you think @erizocosmico @mcarmonaa @jfontan ?

erizocosmico:
Closes #132 

In this PR BaseSession looks really useless, but that's because we're missing the things in #129 and so on. In the future, BaseSession might have a lot more things.


server/context.go
```diff
@@ -0,0 +1,98 @@
...
+
+// SessionBuilder creates sessions given a context and a MySQL connection.
+type SessionBuilder func(*mysql.Conn) sql.Session
+
+// DoneFunc is a function that must be executed when the session is used and
+// it can be disposed.
+type DoneFunc func()
+
+// DefaultSessionBuilder is a SessionBuilder that returns a base session with
+// just the passed context in it.
```
ajnavarro: right now no context is passed.

erizocosmico:
Fixed @ajnavarro

erizocosmico:
Rebased.

PR #35, opened 2018-02-21
sql/expression: implement array_length UDF

erizocosmico:
Closes #7 

Depends on #134

PR #135, opened 2018-03-22
sql: add Array type

### LINKED ISSUE #134: Array type
Implement array type.

erizocosmico:
Closes #134 

**NOTE:** because of the usage of `[]interface{}` internally, there's a side effect of implementing this type and is that arrays can be cast to tuples and vice-versa. Therefore, any function that can use arrays as input can use tuples as input (though that can be enforced inside the functions just as #35 does).

### DIFF
--- a/sql/type.go
+++ b/sql/type.go
@@ -30,9 +30,13 @@ var (
 
 	// ErrNotTuple is retuned when the value is not a tuple.
 	ErrNotTuple = errors.NewKind("value of type %T is not a tuple")
+
 	// ErrInvalidColumnNumber is returned when a tuple has an invalid number of
 	// arguments.
 	ErrInvalidColumnNumber = errors.NewKind("tuple should contain %d column(s), but has %d")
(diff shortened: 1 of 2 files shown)


sql/type.go
```diff
@@ -30,9 +30,13 @@ var (
 
 	// ErrNotTuple is retuned when the value is not a tuple.
 	ErrNotTuple = errors.NewKind("value of type %T is not a tuple")
+
 	// ErrInvalidColumnNumber is returned when a tuple has an invalid number of
 	// arguments.
 	ErrInvalidColumnNumber = errors.NewKind("tuple should contain %d column(s), but has %d")
+
+	// ErrNotArray is returned when the value is not an array.
+	ErrNotArray = errors.NewKind("valud of type %T is not an array")
```
ajnavarro: valud => value

erizocosmico:
Fixed @ajnavarro

PR #136, opened 2018-03-22
sql/expression: fix tuples after merge

erizocosmico:
Fixes the tuples after the merge of the named type


erizocosmico:
If we merge #128 first this PR makes no sense anymore and should be closed.

Issue #137, opened 2018-03-25
Save performance metrics


jfontan:
It would be good to have saved performance metrics that we can use to understand the changes. As mentioned in #36 these metrics can be saved in `Session` or in `Context`.

Some of the data we can save is:

* Time spent executing nodes
* Memory consumption before and after node execution
* Cache hit/miss
* Child nodes to pinpoint slower paths

The proposal is adding a `Performance` interface to base session or context that can be used to save this data. It's done as an interface so we can have several implementations, some can get more info than others or disable collecting data in case it's too expensive.

```go
type Performance interface {
	// Called once per query to reset the counters
	Reset()

	// Marks the start of one node execution, returns an id as there can be several
	// nodes running at the same time
	Start(nodeName, arguments string) (id int)

	// Marks the start of one node execution
	End(id int)

	// This can be used to save other values like cache hit and miss rates
	ExtraData(id int, name string, value int)

	// Gets the collected data in some structured format like json
	GetData() string
}
```

Then all the nodes have to be instrumented to save this data:

```go
func (ib *IsBinary) Eval(
	ctx *sql.Context,
	row sql.Row,
) (interface{}, error) {
	// Start collecting data
	perID := ctx.Performance.Start("IsBinary", ib.String())
	defer ctx.Performance.End(perID)

	v, err := ib.Child.Eval(ctx, row)

	[...]

	blob, err := sql.Blob.Convert(v)
	if err != nil {
		return nil, err
	}

	// Add extra information that may be useful
	ctx.Performance.ExtraData(perID, "blob size", blob.Size())

	return isBinary(blob.([]byte)), nil
}
```

Extra tooling should be created to read and analyze this data. It would also be interesting to automate running tests with old and new code (for example `master` and PR) and comparing the data to find regressions.

Thoughts? @ajnavarro, @erizocosmico, @mcarmonaa

ajnavarro:
I like the idea and it fits on the gitbase backlog. Just a couple of changes:

- The name of the interface should be something like `Metrics`
- The first implementation of that interface should be using open metrics format, on that case using https://github.com/jaegertracing/jaeger

erizocosmico:
Instead of jaeger, we could use opencensus (https://github.com/census-instrumentation/opencensus-go), which has exporters for jaeger and many more, if we want to avoid vendor lock-in

ajnavarro:
@smola , what do you think about use opencensus instead of jaeger? for me is ok.

[...]

PR #138, opened 2018-03-27
*: implement span tracing in sql.Context

erizocosmico:
Part of #137 

This just integrates jaeger into the current `sql.Context`, though it does not yet make use of it.

Basically, which this does is add a new `Span` method to create a new tracing span in `sql.Context`. That gives you the span itself and a new context which is the one that should be passed to the children.
It's basically a wrapper for https://godoc.org/github.com/opentracing/opentracing-go#StartSpanFromContext using the tracer inside the context to make it easier to use inside our current architecture.

Whatever can be passed to the OpenTracing `StartSpan` can be passed to our `Span`.

Also changed the server config to a struct, since it will probably grow over time and the param list would have been a little too huge.


sql/session.go
```diff
@@ -21,9 +25,43 @@ func NewBaseSession() Session {
...
+	return &Context{
+		context.TODO(),
+		NewBaseSession(),
+		opentracing.NoopTracer{},
+	}
+}
+
+// Span creates a new tracing span with the given context.
+// It will return the span and a new context that should be passed to all
+// childrens of this span.
```
jfontan: children

erizocosmico:
Fixed tests, I just added dep to the library. Otherwise jaeger would not build.

PR #140, opened 2018-03-28
 sql/(plan,expression,parse,analyzer): trace node execution 

### LINKED ISSUE #137: Save performance metrics

erizocosmico:
Fixes #137

Depends on #139 and #138 

This commit adds tracing of spans of execution in all nodes and expressions.

Since expressions and functions evaluation is immediate, spans are finished inside the `Eval` function with a defer.
In the case of plan nodes it's a bit more complex, as they just return an iterator that will be lazily evaluated. Measuring the span as just the time it takes to return the iterator would be wrong, so what it does is wrap all iterators with a `spanIter`, which iterates over an internal iterator and finishes the span when it's needed (there is an error, io.EOF is returned or `Close` is called) pushing some metrics like the number of processed rows or the error without needing the user to provide them.

### DIFF
--- a/sql/expression/function/isbinary.go
+++ b/sql/expression/function/isbinary.go
@@ -3,6 +3,10 @@ package function
 import (
 	"bytes"
 	"fmt"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
 
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
@@ -23,6 +27,21 @@ func (ib *IsBinary) Eval(
 	ctx *sql.Context,
 	row sql.Row,
 ) (interface{}, error) {
+	var blobSize int
+	span, ctx := ctx.Span("function.IsBinary")
+	defer func() {
+		span.FinishWithOptions(opentracing.FinishOptions{
+			LogRecords: []opentracing.LogRecord{
+				{
+					Timestamp: time.Now(),
+					Fields: []log.Field{
+						log.Int("blobsize", blobSize),
+					},
+				},
+			},
+		})
+	}()
+
 	v, err := ib.Child.Eval(ctx, row)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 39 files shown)


sql/expression/function/isbinary.go
```diff
@@ -23,6 +27,21 @@ func (ib *IsBinary) Eval(
 	ctx *sql.Context,
 	row sql.Row,
 ) (interface{}, error) {
+	var blobSize int
+	span, ctx := ctx.Span("function.IsBinary")
+	defer func() {
+		span.FinishWithOptions(opentracing.FinishOptions{
```
smola: Soooo verbose :sob:

erizocosmico: I know, the API is a bit eugh 😞

Issue #142, opened 2018-04-02
Support arithmetic operators


ajnavarro:
This query: `SELECT a+b*2+c*3+d*4+e*5, (a+b+c+d+e)/5 FROM t1 ORDER BY 1,2` throws the next error:

```
Received unexpected error:
        			            	unsupported syntax: &sqlparser.BinaryExpr{Operator:"+", Left:(*sqlparser.BinaryExpr)(0xc4203d41e0), Right:(*sqlparser.BinaryExpr)(0xc4203d4210)}
gopkg.in/src-d/go-mysql-server.v0/sql/parse.exprToExpression
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:459
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.selectExprToExpression
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:688
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.selectExprsToExpressions
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:445
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.selectToProjectOrGroupBy
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:415
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.convertSelect
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:99
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.convert
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:56
        			            	gopkg.in/src-d/go-mysql-server.v0/sql/parse.Parse
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/sql/parse/parse.go:46
        			            	gopkg.in/src-d/go-mysql-server%2ev0.(*Engine).Query
        			            		/home/antonio/work/src/gopkg.in/src-d/go-mysql-server.v0/engine.go:30
```

erizocosmico:
No, we don’t have arithmetic operators
[...]

PR #150, opened 2018-04-10
Support arithmetic operators 

kuba--:
https://github.com/src-d/go-mysql-server/issues/142

### DIFF
--- a/sql/expression/arithmetic.go
+++ b/sql/expression/arithmetic.go
@@ -1,0 +1,74 @@
+package expression
+
+import (
+	"fmt"
+
+	"gopkg.in/src-d/go-vitess.v0/vt/sqlparser"
+
+	"github.com/spf13/cast"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// Plus adds two given expressions.
+type Plus struct {
+	BinaryExpression
+}
+
+// NewPlus creates a new Plus sql.Expression.
+func NewPlus(left, right sql.Expression) *Plus {
+	return &Plus{BinaryExpression{Left: left, Right: right}}
+}
+
+func (plus *Plus) String() string {
+	return fmt.Sprintf("%s %s %s", plus.Left, sqlparser.PlusStr, plus.Right)
+}
+
+// Type implements the Expression interface.
+func (*Plus) Type() sql.Type {
+	return sql.Float64
+}
+
+// Eval implements the Expression interface.
+func (plus *Plus) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
+	span, ctx := ctx.Span("expression.Plus")
+	defer span.Finish()
+
+	lval, err := plus.Left.Eval(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+
+	// maybe we should call plus.Type().Convert(sqltypes.Float64) first, but
+	// at the end we have to force casting to float64,
+	// so this is just a shortcut
+	lval64, err := cast.ToFloat64E(lval)
+	if err != nil {
+		return nil, err
+	}
+
+	rval, err := plus.Right.Eval(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+	rval64, err := cast.ToFloat64E(rval)
+	if err != nil {
+		return nil, err
+	}
+
+	return lval64 + rval64, nil
+}
+
+// TransformUp implements the Expression interface.
+func (plus *Plus) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
+	l, err := plus.Left.TransformUp(f)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := plus.Right.TransformUp(f)
+	if err != nil {
+		return nil, err
+	}
+
+	return f(NewPlus(l, r))
+}
--- a/sql/expression/arithmetic.go
+++ b/sql/expression/arithmetic.go
@@ -3,72 +3,295 @@ package expression
 import (
 	"fmt"
 
+	"github.com/spf13/cast"
 	"gopkg.in/src-d/go-vitess.v0/vt/sqlparser"
 
-	"github.com/spf13/cast"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
 
-// Plus adds two given expressions.
-type Plus struct {
+// Arithmetic expressions (+, -, *, /, ...)
+type Arithmetic struct {
 	BinaryExpression
+	op string
 }
 
-// NewPlus creates a new Plus sql.Expression.
-func NewPlus(left, right sql.Expression) *Plus {
-	return &Plus{BinaryExpression{Left: left, Right: right}}
+// NewArithmetic creates a new Arithmetic sql.Expression.
+func NewArithmetic(left, right sql.Expression, op string) *Arithmetic {
+	return &Arithmetic{BinaryExpression{Left: left, Right: right}, op}
 }
 
-func (plus *Plus) String() string {
-	return fmt.Sprintf("%s %s %s", plus.Left, sqlparser.PlusStr, plus.Right)
+// NewPlus creates a new Arithmetic + sql.Expression.
+func NewPlus(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.PlusStr)
+}
+
+// NewMinus creates a new Arithmetic - sql.Expression.
+func NewMinus(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.MinusStr)
+}
+
+// NewMult creates a new Arithmetic * sql.Expression.
+func NewMult(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.MultStr)
+}
+
+// NewDiv creates a new Arithmetic / sql.Expression.
+func NewDiv(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.DivStr)
+}
+
+// NewShiftLeft creates a new Arithmetic << sql.Expression.
+func NewShiftLeft(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.ShiftLeftStr)
+}
+
+// NewShiftRight creates a new Arithmetic >> sql.Expression.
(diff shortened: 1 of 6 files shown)


sql/expression/arithmetic.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// NewArithmetic creates a new Arithmetic sql.Expression.
+func NewArithmetic(left, right sql.Expression, op string) *Arithmetic {
+	return &Arithmetic{BinaryExpression{Left: left, Right: right}, op}
+}
+
+// NewPlus creates a new Arithmetic + sql.Expression.
+func NewPlus(left, right sql.Expression) *Arithmetic {
+	return NewArithmetic(left, right, sqlparser.PlusStr)
```
mcarmonaa: Should we depend on go-vitess for the arithmetics strings or should we have our own constants for this?

kuba--: @mcarmonaa - I had the same thought, but we already rely on `sqlparser` consts (`ast.go`). in `parse.go`, so if we introduce another one we'll have to think about sync it up in case of any potential future changes. So I decided to use the predefined consts. everywhere for consistency.

erizocosmico:
@kuba-- but +, - and * could avoid being casted to float64 when both are integers.

jfontan:
LGTM. As @erizocosmico says I'm not sure about the implications of casing always to float64 even if it's only for those operations.

ajnavarro:
@kuba-- we try to mimic the same logic as MySQL has. On this case here is the types for arithmetic operations: https://dev.mysql.com/doc/refman/5.7/en/arithmetic-functions.html

kuba--:
@erizocosmico, @jfontan, @ajnavarro - OK guys, lets do it in a right way (not just `int64` or `float64`). I've put an effort to create an _lowest common type_ lookup table (https://github.com/src-d/go-mysql-server/pull/150/files#diff-4de3ba69d5dd2df6cef985827947a6f0R1202).

Right now we should return the lowest common type for two values. So if you put two `int8` and result will fit into `int8` then `Eval` will return `int8`.

If we find this _lookup table_ generally useful we can extract it to another file. So far I left it in `arithmetic.go` because it's very specific for arithmetic operations.

kuba--:
...just to recap - if the code is too complicated (even if it's more precise as we needed) then I can get rid of lookup table and just stick to _64-bit_ cases where you have `float64` or `int64`, `int64` for bit operations and `uint64` for shifts. But it means that even if you add `int8` to `int8` we'll return `int64`- WDYT ?

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,1386 @@
...
+}
+
+func bitOr(lval, rval interface{}) (interface{}, error) {
+	kind, err := commonKind(lval, rval)
+	if err != nil {
+		return nil, err
+	}
+
+	switch kind {
+	case reflect.Int:
```
erizocosmico: Yeah, this may be a little too much of an overkill for this. So is using reflection for this, as we already have the types of the operands.

I think is enough to use the following functions: https://github.com/src-d/go-mysql-server/blob/master/sql/type.go#L652-L670 to check which type it is and just treat the following four cases:
- int + int (independently of which size): treat as int64
- uint + uint (independently of which size): treat as uint64
- uint + int (independently of which size): treat as uint64
- anything else (independently of which size): float64

This would be simpler while still correct.

ajnavarro:
Yeah, this task is more difficult than I thought :laughing:

For example, that query `SELECT '3' + 1` returns `4` and `SELECT '' + 4` returns `4` too.

Maybe we need another approach for this.

All this casting logic is already done on `convert.go` so maybe we can wrap both sides of the arithmetic expression with a cast node, casting to the smaller possible type, depending of the both expression types.

What do you think @kuba-- @erizocosmico @jfontan @mcarmonaa ?

erizocosmico:
@ajnavarro sounds good to me

mcarmonaa:
There is something similar done for the comparison expression https://github.com/src-d/go-mysql-server/blob/master/sql/expression/comparison.go#L70

sql/expression/arithmetic.go
kuba--: I used `reflect`, because I needed all basic types. `sql/type.go` are tiny subset, moreover you have to rely on what `Type()` function returns for your left/right what's not always true, because you can pass any value, but your `Type` may return something different (e.g. because in Type you don't want to re-evaluate expressions).

Anyway if we all agree that 64-bit numbers as a result are good enough for all arithmetic operations (and does not matter what's the input's size) then it's fine. I'll refine PR.

kuba--:
mysql also lets you do _the wrong_ things:
```
mysql> select -256 >> 2 as R;
+---------------------+
| R                                        |
+---------------------+
| 4611686018427387840 |
+---------------------+
```

kuba--:
Ok guys, I've simplified/refined the PR. We rely only on 64-bit types and I payed attention that we convert to the same type as `Type` function returns.
@erizocosmico @jfontan @mcarmonaa @ajnavarro  - PTAL

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,411 @@
...
+	"gopkg.in/src-d/go-vitess.v0/vt/sqlparser"
+
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+var (
+	// ErrUnableToCast means that we could not find common type for two arithemtic objects
+	errUnableToCast = errors.NewKind("Unable to cast between types: %T, %T")
+
+	// ErrUnableToEval means that we could not evaluate an expression
```
erizocosmico: s/ErrUnableToEval/errUnableToEval/

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,411 @@
...
+
+	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-vitess.v0/vt/sqlparser"
+
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+var (
+	// ErrUnableToCast means that we could not find common type for two arithemtic objects
+	errUnableToCast = errors.NewKind("Unable to cast between types: %T, %T")
```
erizocosmico: s/ErrUnableToCast/errUnableToCast/

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,411 @@
...
+	case sqlparser.IntDivStr:
+		return intDiv(lval, rval)
+	case sqlparser.ModStr:
+		return mod(lval, rval)
+	}
+
+	return nil, errUnableToEval.New(lval, a.op, rval)
+}
+
+func (a *Arithmetic) eval2(ctx *sql.Context, row sql.Row) (interface{}, interface{}, error) {
```
erizocosmico: why `eval2`?

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,411 @@
...
+
+func (a *Arithmetic) String() string {
+	return fmt.Sprintf("%s %s %s", a.Left, a.op, a.Right)
+}
+
+// Type returns the greatest type for given operation.
+func (a *Arithmetic) Type() sql.Type {
+	switch a.op {
+	case sqlparser.PlusStr, sqlparser.MinusStr, sqlparser.MultStr, sqlparser.DivStr:
+		if sql.IsInteger(a.Left.Type()) && sql.IsInteger(a.Right.Type()) {
```
erizocosmico: what if `a.Left` or `a.Right` is `sql.Text`? `'2' + 3` is a valid expression

sql/expression/arithmetic.go
```diff
@@ -0,0 +1,411 @@
...
+
+	rval, err := a.Right.Eval(ctx, row)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return lval, rval, nil
+}
+
+func (a *Arithmetic) convert2(lval interface{}, rval interface{}) (interface{}, interface{}, error) {
```
erizocosmico: why `convert2`?

sql/expression/arithmetic.go
kuba--: I think we can focus on _number_ right now a extend it by text later. With text you may have different formats like: '1', '1.0', '0x1', '1e-1', etc.

sql/expression/arithmetic.go
kuba--: the same reason like eval2. It's a internal stuff working for both (left and right)

kuba--: it's like a notation `eval/2` (known from many langs.)

kuba--: just to make it less confusing I renamed these function to `evalLeftRight` and `convertLeftRight`

sql/expression/arithmetic.go
kuba--: Actually it should also work for strings. I've double checked and wrote a test:
```
	require := require.New(t)
	result, err := NewPlus(NewLiteral("2", sql.Text), NewLiteral(3, sql.Float64)).
		Eval(sql.NewEmptyContext(), sql.NewRow())
	require.NoError(err)
	require.Equal(float64(5), result)
```
and it works.

And the reason why is because I use:
```
typ := a.Type()
lval64, err := typ.Convert(lval)
```
what internally uses `ToFloat64E`, ... which has a case:
```
	case string:
		v, err := strconv.ParseFloat(s, 64)
		if err == nil {
			return v, nil
		}
```

kuba--: but because upfront we do not know a type of the expression inside a string we can either rely on the second possible type (left or right, but it also can be a string) or just convert everything what's not predefined to float64
Right now if one of expressions is a text we convert result to float64, even if the second _branch_ can be the `int`, because it can be sth. like `'1.0e-1' + 1`

kuba--:
But current implementation works for strings (moreover behaves like mysql). I even added more tests for that (in `parser` and `expression`).
```go
	`SELECT '1.0' + 2;`: plan.NewProject(
		[]sql.Expression{
			expression.NewPlus(
				expression.NewLiteral("1.0", sql.Text), expression.NewLiteral(int64(2), sql.Int64),
			),
		},
		plan.NewUnresolvedTable("dual"),
	),
	`SELECT '1' + '2';`: plan.NewProject(
		[]sql.Expression{
			expression.NewPlus(
				expression.NewLiteral("1", sql.Text), expression.NewLiteral("2", sql.Text),
			),
		},
		plan.NewUnresolvedTable("dual"),
	),
```

PR #152, opened 2018-04-11
Fixes - add missed pointers

kuba--:
I believe all fixes are just typos with missed pointers - PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ _testmain.go
 # Tpc-h tbl files
 benchmark/*tbl
 vendor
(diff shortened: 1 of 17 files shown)


erizocosmico:
They're not typos, it's _intentional_. Since those methods don't make use of anything inside the structs (or only do so for reading), it doesn't matter if it's a pointer receiver or not.

kuba--:
So, shall I close this PR?

ajnavarro:
We can leave all of them as pointers to have consistency between nodes.

Issue #151, opened 2018-04-11
Some changes to opentracing reported metrics


ajnavarro:
To improve the use of opentracing, we should do some minor changes:

- Tags should only be used with data that can be used to filter between operations (per example, on `analyze` operation, we can add a tag with the output of `IsResolved` of the plan). Right now the `query` is set on a tag, move it to a `span.Log()`

- On `analyze` and `analyze_once` add a log with the plan of the node that is being processed.

- Be able to use the default jaeger [environment variables](https://github.com/jaegertracing/jaeger-client-go#environment-variables) instead of custom configuration. To do that, there is a tracer constructor on jaeger `config` package.

- Remove spans from `Merge` and `Update` aggregation methods, they are really verbose and say almost nothing, and add the span to `Eval`.

- Right now, we cannot follow the entire trace of a query because it is splitted into several parts with no relations between them; `parse`, `analyze` and the execution of the query itself. We must be able to trace at query level. 

- The tracer name must be configurable, not hardcoded to `go-mysql-server` (this can be done using `JAEGER_SERVICE_NAME`)

ajnavarro:
What do you think about that @src-d/data-retrieval @smola ?

smola:
@ajnavarro Sounds good. I think the most important part is that we can trace the execution phase. If we were able to track only execution (with a payload with the plan), that would be great. If other phases are there, that's ok.

mcuadros:
I will need to see in real-life, but at the end whatever is useful for you guys.

PR #154, opened 2018-04-16
Opentracing

kuba--:
Description: https://github.com/src-d/go-mysql-server/issues/151

### DIFF
--- a/_example/main.go
+++ b/_example/main.go
@@ -32,7 +32,7 @@ func main() {
 		Password: "password1",
 	}}
 
-	s, err := server.NewDefaultServer("tcp", "localhost:5123", auth, driver)
+	s, err := server.NewDefaultServer(server.Config{"tcp", "localhost:5123", auth}, driver)
 	if err != nil {
 		panic(err)
 	}
@@ -41,7 +41,7 @@ func main() {
 }
 
 func createTestDatabase() *mem.Database {
-	db := mem.NewDatabase("test")
+	db := mem.NewDatabase("test").(*mem.Database)
 	table := mem.NewTable("mytable", sql.Schema{
 		{Name: "name", Type: sql.Text},
 		{Name: "email", Type: sql.Text},
--- a/_example/main.go
+++ b/_example/main.go
@@ -32,21 +32,24 @@ func main() {
 		Password: "password1",
 	}}
 
-	s, err := server.NewDefaultServer(server.Config{"tcp", "localhost:5123", auth}, driver)
-	if err != nil {
+	if s, err := server.NewDefaultServer(server.Config{
+		Protocol: "tcp",
+		Address:  "localhost:5123",
+		Auth:     auth,
+	}, driver); err != nil {
 		panic(err)
+	} else {
+		s.Start()
 	}
-
-	s.Start()
 }
 
 func createTestDatabase() *mem.Database {
 	db := mem.NewDatabase("test").(*mem.Database)
 	table := mem.NewTable("mytable", sql.Schema{
-		{Name: "name", Type: sql.Text},
-		{Name: "email", Type: sql.Text},
-		{Name: "phone_numbers", Type: sql.JSON},
-		{Name: "created_at", Type: sql.Timestamp},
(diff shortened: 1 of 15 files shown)


server/config.go
```diff
@@ -0,0 +1,222 @@
...
+
+	opentracing "github.com/opentracing/opentracing-go"
+	jaeger "github.com/uber/jaeger-client-go"
+	jaegercfg "github.com/uber/jaeger-client-go/config"
+	"github.com/uber/jaeger-lib/metrics"
+	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-vitess.v0/mysql"
+)
+
+const (
```
ajnavarro: You don't need this, you can use https://github.com/jaegertracing/jaeger-client-go/blob/6cc524e2736308fb6da8f054483d00baee39b0a1/config/config_env.go#L49 to get configuration from environment variables.

ajnavarro: And then, with this configuration, you can create a new tracer: https://sourcegraph.com/github.com/jaegertracing/jaeger-client-go@6cc524e2736308fb6da8f054483d00baee39b0a1/-/blob/config/config.go#L148

kuba--: Yeah, I know, but it was added recently and it's still just on master. I hope after they release a new version we'll be able to get rid of some stuff and directly use heir `FromEnv` functions.

ajnavarro: Oh, ok, sorry for the noise. Can you open an issue to remind change that on the future?

_example/main.go
```diff
@@ -32,21 +32,24 @@ func main() {
...
 
-	s, err := server.NewDefaultServer("tcp", "localhost:5123", auth, driver)
-	if err != nil {
+	if s, err := server.NewDefaultServer(server.Config{
+		Protocol: "tcp",
+		Address:  "localhost:5123",
+		Auth:     auth,
+	}, driver); err != nil {
 		panic(err)
+	} else {
```
erizocosmico: no need for an else, panic stops execution

_example/main.go
```diff
@@ -32,21 +32,24 @@ func main() {
...
+		s.Start()
 	}
-
-	s.Start()
 }
 
 func createTestDatabase() *mem.Database {
-	db := mem.NewDatabase("test")
+	db := mem.NewDatabase("test").(*mem.Database)
```
erizocosmico: why is the casting needed?

server/config.go
kuba--: Wait, I've just checked that they released a new version `2.13` two days ago (with env config), so on Friday the latest version was 2.12 (the version which we are using right now).
So, shall we integrate a new version?

sql/expression/function/aggregation/count.go
```diff
@@ -82,14 +79,16 @@ func (c *Count) Update(ctx *sql.Context, buffer, row sql.Row) error {
...
 	buffer[0] = buffer[0].(int32) + partial[0].(int32)
 	return nil
 }
 
 // Eval implements the Aggregation interface.
 func (c *Count) Eval(ctx *sql.Context, buffer sql.Row) (interface{}, error) {
-	return buffer[0], nil
+	span, ctx := ctx.Span("aggregation.Count_Eval")
+	count := buffer[0]
+	span.LogKV("count", count)
```
erizocosmico: why do we need the result of the aggregation in the log?

_example/main.go
kuba--: because `mem.NewDatabase` returns `sql.Database` interface which does not have `AddTable` function.

sql/expression/function/aggregation/count.go
kuba--: ...just to be consistent with other operations (max, min, avg, ...) if it's too noisy we can remove it.

erizocosmico: Yeah, perhaps having in the logs the results of these operations is too noisy

kuba--:
PTAL: jagger-client-go was upgraded to 2.13.0

PR #155, opened 2018-04-17
sql/analyzer: resolve DUAL table

### LINKED ISSUE #153: Register a DUAL table by default
The [DUAL table](https://en.wikipedia.org/wiki/DUAL_table) present [in MySQL](https://dev.mysql.com/doc/refman/5.7/en/select.html) is often used by 3rd party tools, connection pools, etc.

Registering it by default should help us to improve compatibility our of the box. It could be implemented easily with a `mem.Table`.

erizocosmico:
Fixes #153

DUAL is a special table that is used when no other table is. For example, `SELECT 1+2` becomes `SELECT 1+2 FROM DUAL`. Because it is a special table, it does not show up in `SHOW TABLES` or anything like that. For that reason, having the analyzer to resolve that table without needing that it is on the catalog means we don't have to add special checks for this case in the logic that finds tables in the catalog.

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-mysql-server.v0/mem"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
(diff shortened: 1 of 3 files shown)


sql/analyzer/rules.go
```diff
@@ -218,6 +219,14 @@ func resolveDatabase(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error
 	return n, nil
 }
 
+var dualTable = func() sql.Table {
+	t := mem.NewTable("dual", sql.Schema{
+		{Name: "dummy", Source: "dual", Type: sql.Text, Nullable: false},
+	})
+	_ = t.Insert(sql.NewRow("x"))
```
mcarmonaa: Why is necessary insert something in the dual table?

smola: `DUAL` is a single-row, single-column table.

PR #158, opened 2018-04-18
sql/analyzer: resolve columns from the projection down the tree

### LINKED ISSUE #241: Create Pilosa Index - Performance Improvements
So far creating an index is really slow and consumes lot of resources. Here are some numbers:
- I've created a _language index_ based on *4 our repositories*: baremetal, borges, envconfig, gcfg, gitbase.
- 1613800 rows saved
- It took ~ *2 hours*
- pilosa-mapping.db file size ~ *824 MB*

erizocosmico:
Fixes https://github.com/src-d/gitbase/issues/241

This commit introduces 2 new analyzer rules:

- `erase_projection`, which deletes redundant `Project` nodes. When a `Project`s schema is exactly the same as the schema of its child, the `Project` node is removed. For example `SELECT * FROM table` would be something like `Project -> Table`, which is converted to only `Table` now because the `Project` node is redundant.
- `reorder_projection`, which moves projected columns down the tree so that appearances of those columns in sort and filter nodes may be resolved. The way we parse the queries makes the `Project` the topmost node of the tree most of the time and `Filter` and `Sort` are always down `Project`. That makes it impossible for the analyzer to resolve the columns that are _created_ in the projection (aliases). Now, those columns are pushed down below the node that requires them in a `Project` node.

There also have been some modifications to current analyzer rules:

- `qualify_columns` now does not error if the column can't be qualified. Since we're gonna need resolution for aliases and aliases don't belong to any table, having columns that cannot be qualified is expected.
- `resolve_columns` now does not error if it fails to resolve a column in it's first pass. That's so other rules may have time to do some work on the tree so that this column can be resolved. Instead, that first pass the column is wrapped with `maybeAlias` and then in any subsequent pass if we find a `maybeAlias` and we can't resolve it this time it means the other analyzer rules weren't able to make changes in the tree to make this column resolvable and it fails with an error. This is a way to defer the column resolution because some rules may need to be able to use the schema (which requires all the nodes with columns down the tree resolved) in order to do some work and make some columns resolvable (such as `reorder_projection`).

Other small changes:

- `plan.Project` now also propagates the `Source` of the column if the expression had it, which it did not before.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -17,209 +17,192 @@ import (
 
 const driverName = "engine_tests"
 
-func TestQueries(t *testing.T) {
-	e := newEngine(t)
-
-	testQuery(t, e,
+var queries = []struct {
+	query    string
+	expected []sql.Row
+}{
+	{
 		"SELECT i FROM mytable;",
-		[][]interface{}{{int64(1)}, {int64(2)}, {int64(3)}},
-	)
-
-	testQuery(t, e,
+		[]sql.Row{{int64(1)}, {int64(2)}, {int64(3)}},
+	},
+	{
 		"SELECT i FROM mytable WHERE i = 2;",
-		[][]interface{}{{int64(2)}},
-	)
-
-	testQuery(t, e,
+		[]sql.Row{{int64(2)}},
+	},
+	{
 		"SELECT i FROM mytable ORDER BY i DESC;",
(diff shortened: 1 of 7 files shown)


mcarmonaa:
I guess the issue this PR fixes comes from gitbase, right? https://github.com/src-d/gitbase/issues/241

erizocosmico:
@mcarmonaa yep

PR #166, opened 2018-04-24
server: remove specific tracer implementation

### LINKED ISSUE #165: go dep
This package is a library, and as a library shouldn't be using `go dep`, if vendor is required for certain packages, then only this packages should be vendored, and not the other dependencies.

And I bless God to not having anything vendored, in the near future.

erizocosmico:
Closes #165 

That means no more dep on this library.


ajnavarro:
We need to remove dep calls from travis.yaml to be able to execute the tests.

PR #179, opened 2018-04-27
sql: implement index catalog

### LINKED ISSUE #170: General IndexCatalog implementation

erizocosmico:
Closes #170 

I removed the `Status() IndexStatus` from the `Index` interface, since doing that would require a `SetStatus` and I don't really think we should expose that method.

### DIFF
--- a/sql/index.go
+++ b/sql/index.go
@@ -1,6 +1,14 @@
 package sql
 
-import "io"
+import (
+	"context"
+	"io"
+	"reflect"
+	"strings"
+	"sync"
+
+	"gopkg.in/src-d/go-errors.v1"
+)
 
 // IndexKeyValueIter is an iterator of index key values, that is, a tuple of
 // the values that will be index keys.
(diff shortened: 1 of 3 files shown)


sql/index.go
```diff
@@ -87,18 +99,272 @@ type Mergeable interface {
...
+}
+
+// Index returns the index with the given id. It may return nil if the index is
+// not found.
+func (r *IndexRegistry) Index(db, id string) Index {
+	r.mut.RLock()
+	defer r.mut.RUnlock()
+
+	id = strings.ToLower(id)
+	for _, idx := range r.indexes {
```
kuba--: maybe `indexes` can be a `map[db.id]Index` (if `id` is unique per db) so we can speed up a lookup

erizocosmico: We need to iterate anyway for the `IndexByExpression`, but ok, we can speed up getting by key.

erizocosmico:
Done @kuba--

ajnavarro:
@kuba-- @erizocosmico shall we merge this?

erizocosmico:
SGTM

PR #181, opened 2018-05-04
sql/analyzer: resolve correctly GetField expressions indexes

### LINKED ISSUE #180: Row's indexes are not handled properly
For queries involving `CROSS JOIN` and `INNER JOIN` the column resolution is done right for the resulting schema, but is not handle correctly for intermediate phases of the execution.

For example, for a query like:
```
SELECT * FROM repositories r,
		commits c INNER JOIN tree_entries te
			ON c.tree_hash = te.tree_hash
```

with the following plan:
```
CrossJoin
 ├─ TableAlias(r)
 │   └─ PushdownProjectionAndFiltersTable
 │       ├─ Columns(repositories.id)
 │       ├─ Filters()
 │       └─ Table(repositories)
 │           └─ Column(id, TEXT, nullable=false)
 └─ InnerJoin(commits.tree_hash = tree_entries.tree_hash)
     ├─ TableAlias(c)
     │   └─ PushdownProjectionAndFiltersTable
     │       ├─ Columns(commits.hash, commits.author_name, commits.author_email, commits.author_when, commits.committer_name, commits.committer_email, commits.committer_when, commits.message, commits.tree_hash, commits.parents)
     │       ├─ Filters()
     │       └─ Table(commits)
     │           ├─ Column(hash, TEXT, nullable=false)
     │           ├─ Column(author_name, TEXT, nullable=false)
     │           ├─ Column(author_email, TEXT, nullable=false)
     │           ├─ Column(author_when, TIMESTAMP, nullable=false)
     │           ├─ Column(committer_name, TEXT, nullable=false)
     │           ├─ Column(committer_email, TEXT, nullable=false)
     │           ├─ Column(committer_when, TIMESTAMP, nullable=false)
     │           ├─ Column(message, TEXT, nullable=false)
     │           ├─ Column(tree_hash, TEXT, nullable=false)
     │           └─ Column(parents, JSON, nullable=false)
     └─ TableAlias(te)
         └─ PushdownProjectionAndFiltersTable
             ├─ Columns(tree_entries.tree_hash, tree_entries.entry_hash, tree_entries.mode, tree_entries.name)
             ├─ Filters()
             └─ Table(tree_entries)
                 ├─ Column(tree_hash, TEXT, nullable=false)
                 ├─ Column(entry_hash, TEXT, nullable=false)
                 ├─ Column(mode, TEXT, nullable=false)
                 └─ Column(name, TEXT, nullable=false)

```

The resulting schema is:
```
0: id
1: hash
2: author_name
3: author_email
4: author_when
5: committer_name
6: committer_email
7: committer_when
8: message
9: tree_hash
10: parents
11: tree_hash
12: entry_hash
13: mode
14: name
```
When the `resolveColumns` rule in the `analyzer` changes the `GetField` expressions in the plan, it set the row's index where the value for each field of the schema should be placed in a `sql.Row` referred to this schema.

However, for the query example, when the rows are evaluated in the inner join iterator, the value for first field `id` that comes from the `repository` table, isn't present in the `sql.Row` at this point because the cross join hasn't produced yet.

 Then, the condition for the inner join that use the fields `tree_hash` will take the indexes set in the `GetField` expression 9 and 11, but actually the the `sql.Row` in this iterator only has 14 values, not 15, and the correct fields are in the positions 8 and 10.

In this example, when you try to iterate to get the rows it panics because of the wrong selected values to perform the join have different types and can't be compared. In other cases it could be panic with an `index out of bounds`, or in the worst case, it could not panic and make the inner join on wrong fields.

mcarmonaa:
Fixes #180 

Signed-off-by: Manuel Carmona <manu.carmona90@gmail.com>

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -196,6 +196,20 @@ var queries = []struct {
 		`SELECT i AS foo FROM mytable WHERE foo NOT IN (1, 2, 5)`,
 		[]sql.Row{{int64(3)}},
 	},
+	{
+		`SELECT * FROM tabletest, mytable mt INNER JOIN othertable ot ON mt.i = ot.i2`,
+		[]sql.Row{
+			{"a", int32(1), int64(1), "first row", "third", int64(1)},
+			{"a", int32(1), int64(2), "second row", "second", int64(2)},
+			{"a", int32(1), int64(3), "third row", "first", int64(3)},
+			{"b", int32(2), int64(1), "first row", "third", int64(1)},
+			{"b", int32(2), int64(2), "second row", "second", int64(2)},
+			{"b", int32(2), int64(3), "third row", "first", int64(3)},
+			{"c", int32(3), int64(1), "first row", "third", int64(1)},
+			{"c", int32(3), int64(2), "second row", "second", int64(2)},
+			{"c", int32(3), int64(3), "third row", "first", int64(3)},
+		},
+	},
 }
 
 func TestQueries(t *testing.T) {
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -308,11 +308,6 @@ func resolveStar(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 	})
 }
 
-type columnInfo struct {
-	idx int
-	col *sql.Column
-}
-
 // maybeAlias is a wrapper on UnresolvedColumn used only to defer the
 // resolution of the column because it could be an alias and that
 // phase of the analyzer has not run yet.
@@ -342,20 +337,23 @@ func resolveColumns(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error)
 			return n, nil
 		}
 
-		colMap := make(map[string][]columnInfo)
-		idx := 0
+		colMap := make(map[string][]*sql.Column)
 		for _, child := range n.Children() {
 			if !child.Resolved() {
 				return n, nil
 			}
 
 			for _, col := range child.Schema() {
-				colMap[col.Name] = append(colMap[col.Name], columnInfo{idx, col})
-				idx++
+				colMap[col.Name] = append(colMap[col.Name], col)
 			}
 		}
 
-		return n.TransformExpressionsUp(func(e sql.Expression) (sql.Expression, error) {
+		expressioner, ok := n.(sql.Expressioner)
+		if !ok {
+			return n, nil
+		}
+
+		return expressioner.TransformExpressions(func(e sql.Expression) (sql.Expression, error) {
 			a.Log("transforming expression of type: %T", e)
 			if n.Resolved() {
 				return e, nil
@@ -366,7 +364,7 @@ func resolveColumns(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error)
 				return e, nil
 			}
 
-			columnsInfo, ok := colMap[uc.Name()]
+			columns, ok := colMap[uc.Name()]
 			if !ok {
 				if uc.Table() != "" {
 					return nil, ErrColumnTableNotFound.New(uc.Table(), uc.Name())
@@ -380,11 +378,11 @@ func resolveColumns(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error)
 				}
 			}
 
-			var ci columnInfo
+			var col *sql.Column
 			var found bool
-			for _, c := range columnsInfo {
-				if c.col.Source == uc.Table() {
-					ci = c
+			for _, c := range columns {
+				if c.Source == uc.Table() {
+					col = c
 					found = true
 					break
 				}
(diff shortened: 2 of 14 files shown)


mcarmonaa:
It seems st is failing, don't review yet

mcarmonaa:
@ajnavarro now CI passes, it's ready to review!

engine_test.go
```diff
@@ -206,6 +206,32 @@ func TestQueries(t *testing.T) {
 	}
 }
 
+func TestJoinsFields(t *testing.T) {
```
ajnavarro: Can you add the query to the `queries` slice instead of create a new test for this?

sql/analyzer/rules.go
```diff
@@ -809,3 +810,49 @@ func fixFieldIndexes(schema sql.Schema, exp sql.Expression) (sql.Expression, err
 		return e, nil
 	})
 }
+
+func fixJoinFields(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
```
ajnavarro: Can you add some comments to this function explaining why this is needed and adding a TODO to fix it in a proper way?

sql/analyzer/rules.go
```diff
@@ -809,3 +810,49 @@ func fixFieldIndexes(schema sql.Schema, exp sql.Expression) (sql.Expression, err
...
+	defer span.Finish()
+
+	a.Log("fix join fields, node of type: %T", node)
+	return node.TransformUp(func(n sql.Node) (sql.Node, error) {
+		a.Log("transforming node of type: %T", n)
+		if !n.Resolved() {
+			return n, nil
+		}
+
+		_, ok := n.(*plan.InnerJoin)
```
ajnavarro: Only for inner joins? should be for all kind of joins, right?

mcarmonaa:
@ajnavarro @erizocosmico 

I changed the PR to address all cases. Now there is a rule `resolveGetFieldIndexes` in charge of set the correct index for the `GetField` expressions referencing this value to the `sql.Node` that contains that expression.

Thus, the `resolveColumns` rule doesn't set these indexes anymore and we avoid the usage of the global row's indexes based on the final schema.

sql/analyzer/rules.go
```diff
@@ -330,6 +326,8 @@ type column interface {
 	sql.Expression
 }
 
+const unresolvedGetFieldIndex = 2147483647
```
erizocosmico: -1 would do the trick and then you can do `index >= 0` to know if it's resolved

sql/plan/innerjoin.go
```diff
@@ -106,14 +106,25 @@ func (j *InnerJoin) TransformExpressionsUp(f sql.TransformExprFunc) (sql.Node, e
...
 }
 
 // Expressions implements the Expressioner interface.
-func (j InnerJoin) Expressions() []sql.Expression {
+func (j *InnerJoin) Expressions() []sql.Expression {
 	return []sql.Expression{j.Cond}
 }
+
+// TransformExpressions implements the Expressioner interface.
+func (j *InnerJoin) TransformExpressions(f sql.TransformExprFunc) error {
```
erizocosmico: why add this method when nodes already have `TransformExpressionsUp`?

mcarmonaa: Because `TransformExpressionsUp` transforms expressions for this node and its childrens meanwhile this method only transform the expressions for this node.

erizocosmico: Ok. In any case, all nodes should be immutables, so they would need to return new instances, just as `TransformExpressionsUp` and `TransformUp` do.

ajnavarro:
@mcarmonaa You can do the same logic calling `Expressions()` method on the nodes that you need into a rule. I cannot see the need of `TransformExpression()` method.

mcarmonaa:
@ajnavarro, calling `Expressions` you will get the expressions for this node, but you must add a different implementation to generate a new node with its expressions changed for each node subtype(`inner join`, `project`, etc) in the rule.

Having the method `TransformExpressions` you have that specific implementation done in each node, which I think it's better.

ajnavarro:
Yes, but you can do it in a rule checking the type of the node depending on the case.

ajnavarro:
@erizocosmico can you have a look?

sql/type.go
```diff
@@ -63,14 +63,15 @@ func (s Schema) CheckRow(row Row) error {
...
 }
 
-// Contains returns whether the schema contains a column with the given name.
-func (s Schema) Contains(column string) bool {
-	for _, col := range s {
-		if col.Name == column {
-			return true
+// Contains returns whether the schema contains a column with the given name
+// and the index of that column in the schema.
+func (s Schema) Contains(column string, source string) (int, bool) {
```
erizocosmico: Instead of making contains return an int, which feels weird, maybe we could have a `ColumnIndex` or `IndexOf` method that returns the index or `-1` when it's not found and the following `Contains`:

```go
func (s Schema) Contains(column string, source string) bool {
        return s.IndexOf(column, source) >= 0
}
```

WDYT?

sql/core.go
```diff
@@ -110,6 +110,10 @@ type Node interface {
 type Expressioner interface {
 	// Expressions returns the list of expressions contained by the node.
 	Expressions() []Expression
+	// TransformExprssions applies for each expression in this node
```
erizocosmico: s/TransformExprssions/TransformExpressions/

eiso: Sounds like something that an intelligent code review bot should be able to detect, [hint](https://github.com/src-d/feature-idea/issues/new?template=Custom.md)

ajnavarro:
@erizocosmico ping

ajnavarro:
@erizocosmico ping?

PR #182, opened 2018-05-04
 sql/*: add support for create index statement 

### LINKED ISSUE #171: CREATE INDEX statement
- Implement custom parser to be able to index expressions too
- Add a rule to check if the expression is valid and indexable
- Logic to include the index in the catalog (usage of the IndexLoader) but mark it as disabled until the indexation finish.

erizocosmico:
Depends on #179 

Closes #171 

Missing:

- Parse driver name
- Parse config

There are a few questions around that in the document, they will be implemented once they are answered.


sql/index.go
```diff
@@ -40,8 +40,8 @@ type Index interface {
 	Database() string
 	// Table returns the table name this index belongs to.
 	Table() string
-	// Expression returns the indexed expression.
-	Expression() Expression
+	// Expressions returns the indexed expressions.
```
ajnavarro: Can you add here all the explanation about if there is more than one expression they are Columns and so on?

sql/index.go
```diff
@@ -104,8 +104,8 @@ type Mergeable interface {
 type IndexDriver interface {
 	// ID returns the unique name of the driver.
 	ID() string
-	// Create a new index for the given expression and id.
-	Create(path, db, id string, expr Expression) (Index, error)
+	// Create a new index.
```
ajnavarro: The same as the previous comment here.

sql/index.go
```diff
@@ -120,7 +120,7 @@ type indexKey struct {
 
 // IndexRegistry keeps track of all indexes in the engine.
 type IndexRegistry struct {
-	root string
+	Root string
```
ajnavarro: Some documentation for this, now that is public?

sql/index.go
```diff
@@ -210,11 +210,16 @@ func (r *IndexRegistry) IndexByExpression(db string, expr Expression) Index {
 	defer r.mut.RUnlock()
 
 	for _, idx := range r.indexes {
-		if reflect.DeepEqual(idx.Expression(), expr) && idx.Database() == db {
-			r.retainIndex(db, idx.ID())
-			return idx
+		if idx.Database() == db {
+			for _, e := range idx.Expressions() {
```
ajnavarro: All the expressions must match, not only one. So maybe we need to change ` IndexByExpression(db string, expr Expression) Index` to ` IndexByExpression(db string, expr ...Expression) Index` and use `exprListsEqual`

erizocosmico:
Done @ajnavarro

erizocosmico:
Don't merge this until the `USING` and the config things are implemented.

erizocosmico:
Now it's fully finished. Can be reviewed again.

sql/parse/create_index.go
```diff
@@ -63,10 +73,128 @@ func parseCreateIndex(s string) (sql.Node, error) {
...
+			if err == io.EOF {
+				break
+			}
+
+			if err != nil {
+				return err
+			}
+
+			if singleQuote || doubleQuote {
+				switch true {
```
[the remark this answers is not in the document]
erizocosmico: In this case, what `ignoreNext` does is ignore the meaning of the next character, but we need everything written, that's why there is no `break` in there. We need every character written because we delegate the real parsing to `sqlparser`.

PR #183, opened 2018-05-04
 sql/*: implement drop index statement

### LINKED ISSUE #172: DROP INDEX statement
- Get the id from the statement and add all the logic to find the index from the index catalog
- Add logic to remove the index from the index (disabled) catalog avoiding queries fails if they are using it

erizocosmico:
Depends on #182 

Closes #172


ajnavarro:
could you rebase this one too @erizocosmico ?

erizocosmico:
Sure

erizocosmico:
Fixed @ajnavarro

sql/index.go
```diff
@@ -206,6 +209,8 @@ func (r *IndexRegistry) ReleaseIndex(idx Index) {
 func (r *IndexRegistry) Index(db, id string) Index {
 	r.mut.RLock()
 	defer r.mut.RUnlock()
+	r.retainIndex(db, id)
+	fmt.Println(indexKey{db, strings.ToLower(id)})
```
ajnavarro: leftover?

sql/plan/drop_index.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	driver := d.Catalog.IndexDriver(index.Driver())
+	if driver == nil {
+		return nil, ErrInvalidIndexDriver.New(index.Driver())
+	}
+
+	go func() {
```
ajnavarro: should we block until the index is deleted?

erizocosmico: It depends. Do we want to make the user wait until it's deleted or succeed and asynchronously delete the index?

sql/index.go
erizocosmico: Fixed

sql/plan/drop_index.go
ajnavarro: on index deletion (that shouldn't take A LOT) I think we should wait until the index is dropped, or throw an error otherwise. WDYT @kuba-- ?

kuba--: delete index is/should be super fast (basically it deletes directory). so far it also sends delete to pilosa synchronously (because i didn't notice any lag), but If we encounter any issues we can async. call pilosa). Just to recap, it's totally fine to delete just directory. Pilosa bitmap can be deleted any time or even overwritten.

kuba--: yeah, lets block and wait if we get an error

kuba--: yeah, lets block and wait if we get an error

erizocosmico:
Done @ajnavarro @kuba--

erizocosmico:
Rebased. Any else left so we can merge this, @kuba--?

kuba--:
LGTM

PR #191, opened 2018-05-11
*: implement natural join

### LINKED ISSUE #190: Support NATURAL INNER JOIN statements
Implement Natural inner Join. basically is a standard inner join but instead of providing ON expressions, all the columns with the same names will be used to make the Join.

Also, the Star from the projection will be resolved in a special way: **common columns, left columns, right columns**.

Create a new rule to solve the NATURAL JOINS. Vitess parser has natural joins as totally different joins. We should resolve “natural inner join”.

The natural joins should be converted to his homonym. Per example, natural inner join will be parsed to an inner join on the logical plan, resolving as described before the projection and the conditions depending on the table names.

More info here: https://dev.mysql.com/doc/refman/8.0/en/join.html

erizocosmico:
Closes #190


sql/analyzer/rules.go
```diff
@@ -264,6 +265,103 @@ func resolveTables(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error)
...
+		if !join.Left.Resolved() || !join.Right.Resolved() {
+			return n, nil
+		}
+
+		leftSchema, rightSchema := join.Left.Schema(), join.Right.Schema()
+
+		var conditions, common, left, right []sql.Expression
+		var seen = make(map[string]struct{})
+
+		for i, lcol := range leftSchema {
```
kuba--: This is just a proposal to consider:
Maybe instead O(n^2) iteration here, we can:
- go through the left and store everything in `map[column_name] = expression.NewGetFieldWithTable`
- go through the right and do the same as for left but if `key/column_name` exists then overwrite it.
at the end map should contain all dedup columns in O(2*n), but maybe I'm wrong and missed something, so  it's up to you.

engine_test.go
```diff
@@ -319,6 +319,50 @@ func TestDDL(t *testing.T) {
 	require.Equal(s, testTable.Schema())
 }
 
+func TestNaturalJoin(t *testing.T) {
```
smola: We should test some more cases:

1. NATURAL JOIN on two tables with no common column at all.
2. NATURAL JOIN on two tables with a common column, some rows match the join condition, but some others do not.

sql/analyzer/rules_test.go
```diff
@@ -159,6 +159,55 @@ func TestResolveTablesNested(t *testing.T) {
 	require.Equal(expected, analyzed)
 }
 
+func TestResolveNaturalJoins(t *testing.T) {
+	require := require.New(t)
```
smola: Test resolve on tables with completely disjoint columns. Also with all equal columns.

erizocosmico:
Done @smola

PR #192, opened 2018-05-14
all: add vanity import specs

sbinet:
This CL adds vanity imports to all packages to prevent importing any
given package from the wrong import path.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -1,4 +1,4 @@
-package sqle
+package sqle // import "gopkg.in/src-d/go-mysql-server.v0"
 
 import (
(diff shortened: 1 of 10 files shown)


erizocosmico:
Thanks for the contribution @sbinet! We need all the commits signed off. You have more information here:

https://github.com/probot/dco#how-it-works

You can do that easily for this contribution with:

```
$ git commit --amend -s
$ git push --force
```

sbinet:
PTAL

ajnavarro:
@sbinet thanks a lot!

Issue #194, opened 2018-05-17
Make `sql.Expression` interface de/serializable...


kuba--:
In `IndexDriver` implementation, we have to deal with `sql.Expression` interface (serialize it in `Create` and deserialize in `Load`). Deserialize an interface in *go* is not really possible, so here are couple of proposals how to make `sql.Expression` type more flexible to work with:
- Add *(Un)Marshal* functions, so everyone who implements the interface will have to know how to serialize it to `[]byte` and how to deserialize it from `[]byte`.
- Create `ExpressionRegistry` and all concrete implementations will have to register `Expression` type with unique name in global registry. And downstream this information to components like `IndexDriver` where information about concrete implementation is needed.
- Attach some kind of _hash reference_ to every `sql.Expression` implementation and in components like `IndexDriver`, `sql.Index`, ... use only _hash reference_ strings instead of _expression objects_.

Right now `sql.Expression` interface is:
```go

// Expression is a combination of one or more SQL expressions.
type Expression interface {
	Resolvable
	fmt.Stringer
	// Type returns the expression type.
	Type() Type
	// IsNullable returns whether the expression can be null.
	IsNullable() bool
	// Eval evaluates the given row and returns a result.
	Eval(*Context, Row) (interface{}, error)
	// TransformUp transforms the expression and all its children with the
	// given transform function.
	TransformUp(TransformExprFunc) (Expression, error)
	// Children returns the children expressions of this expression.
	Children() []Expression
}
```

For instance in _Pilosa driver_ apart from (de)serialize it in `Load, Create` we also iterate through all expression in `Save` (because we treat them as column names) and map them to _Pilosa frame names_. To make it possible we just convert them to `string` just calling `String()` function:
```go
for i, e := range idx.Expressions() {
		encodedFrameName := encode(e.String())
		frames[i], err = index.Frame(encodedFrameName)
	...
```

So for _index driver_ it would be better if expressions were just a slice of strings.
This task should be accomplished before we start continue working on: https://github.com/src-d/go-mysql-server/issues/174

ajnavarro:
Should we abstract all this logic from Index implementations, and just send ExpressionsIDs?

Per example:

```go
type Index interface {
...
Expressions() []Expression
...
}
```

Should be:
```go
type ExpressionID string

type Index interface {
...
ExpressionIDs() []ExpressionID
...
}
```

And in the case of `IndexDriver` interface, we can do the same.

[...]

PR #200, opened 2018-05-21
Replace expressions by hashes in IndexDriver

kuba--:
It's hard to deal with `[]sql.Expression` in `IndexDriver` (due to interfaces deserialization issues in go). After meeting with @ajnavarro and @erizocosmico we decided to introduce some kind of `ID` which let us compare two expressions. In `IndexDriver` we do not need _full Expressions_, IDs are good enough. 
This PR introduce a new function `NewExpressionHash(Expression) hash.Hash` and `IndexDriver.ExpressionHashes() []hash.Hash`. A default implementation returns a _sha1_ (but we can replace it by any hash algorithm) based on `Expression.String()`. It would be also easy to map it, because all DBs use _key-value pairs_ as  `[]byte`.

It closes: https://github.com/src-d/go-mysql-server/issues/194

### DIFF
--- a/sql/core.go
+++ b/sql/core.go
@@ -1,7 +1,9 @@
 package sql
 
 import (
+	"crypto/sha1"
 	"fmt"
+	"hash"
 
 	"gopkg.in/src-d/go-errors.v1"
 )
@@ -75,6 +77,14 @@ type Expression interface {
 	TransformUp(TransformExprFunc) (Expression, error)
 	// Children returns the children expressions of this expression.
 	Children() []Expression
+}
+
+// NewExpressionHash returns a new Hash for given Expression instance.
+// Hash (sha1) checksum will be calculated based on ex.String().
+func NewExpressionHash(ex Expression) hash.Hash {
+	h := sha1.New()
+	h.Write([]byte(ex.String()))
+	return h
 }
 
 // Aggregation implements an aggregation expression, where an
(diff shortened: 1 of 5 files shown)


smola:
Makes sense.
I am a bit worried about implementing stuff like `DESCRIBE INDEX` in the future though. When we need it, maybe we can just use SQL as a serialization format.

kuba--:
@smola at the end we have to convert `Expression` to "something" which is max 64 chars. len. and alphanum (because of Pilosa naming validation rules). But this logic can be a part of mapping.

If we need encoding/decoding we can just change a default implementation and replace sha1 hash by base64. I'm totally Ok with this (and see decoding benefits), the only drawback which comes to my mind is that we'll have to go through mapping because of unpredictable base64 string lengths, where for hashes we have fixed length byte slices so we can blindly pass to KV and/or to Pilosa, as well.

erizocosmico:
For the `describe index` we can always store a `expr.String()` and just print it, though

erizocosmico:
LGTM when CI passes.

smola:
Yes, we could even store that SQL separately. Let's defer that until we actually want `DESCRIBE` ;)

kuba--:
CI passed.
Btw. a nice thing about hashes (but I do not know if it's our case) is that you can combine two expression hashes and get a new, unique, fixed length hash which represents the combination.

But if it's too heavy we can compare expressions just by strings.

Issue #174, opened 2018-04-26
Implement Pilosa IndexDriver


ajnavarro:
Create the pilosa IndexDriver implementation:

```go
// IndexDriver manages the coordination between the indexes and their
// representation in disk.
type IndexDriver interface {
	// ID returns the unique name of the driver.
	ID() string
	// Create a new index.
	Create(path, table, db, id string, exprs []Expression) (Index, error)
	// Load the index at the given path.
	Load(path string) (Index, error)
	// Save the given index at the given path.
	Save(ctx context.Context, path string, index Index, iter IndexKeyValueIter) error
	// Delete the index with the given path.
	Delete(path string, index Index) error
}
```

And implement Create and Save Methods.

kuba--:
# Indexing, Pilosa, BadgerDB, Configuration
* Index Driver Interface:
```go

// Create an instance of pilosa driver
func NewDriver(path string, client *pilosa.Client) *Driver {
	return &Driver{
		path:   path,
		client: client,
	}
}

// Pilosa driver implements following interface
type IndexDriver interface {
	ID() string

	Create(db, table, id string, expr []Expression, config map[string]string) (Index, error)

	Load(db, table, id string) (Index, error)

	Save(ctx context.Context, idx Index, iter IndexKeyValueIter) error

	Delete(index Index) error
}
```

* Index directory
```sh
/indexes
└── db_name
    └── table_name
        └── index_id # Configuration and mapping files 
            ├── 000000.vlog 
            ├── 000002.sst 
            ├── MANIFEST # BadgerDB manifest
            └── config.yml


# config.yml
db: mydb
table: mytable
id: myid
drivers:
  pilosa:
    host: localhost
    port: 10101
```

* Pilosa model
![](https://www.pilosa.com/img/docs/data-model.svg)

* Examples
	* `CREATE INDEX test_index ON tree_entries (name);`

		* Table name: tree_entries
		* ID: test_index
		* Expressions: name

		- `Create`
			- Create a folder and config file
			  `/indexes/db/tree_entries/test_index/config.yml` 

		* `Save(idx sql.Index, iter sql.IndexKeyValueIter)`
			* Open BadgerDB with mappings `value` -> `columnID`
			* Get (or Create a new one) pilosa Index (`db_tree_entries_test_index`)
			* For each expression `ex`  in [`name`]
				* Get (or Create a new one) pilosa Index Frame (`$ex`)
				* For each (`key`, `value`) in `iter`
[...]

PR #205, opened 2018-05-25
Implementation of Pilosa Index Driver

kuba--:
It closes: https://github.com/src-d/go-mysql-server/issues/174
Design doc.: https://docs.google.com/document/d/1rWeR739rZsSnCBsXqyviVss8JAO6msV9B1PoDD45O2A/edit?usp=sharing

### DIFF
--- a/sql/index.go
+++ b/sql/index.go
@@ -2,6 +2,7 @@ package sql
 
 import (
 	"context"
+	"encoding/hex"
 	"io"
 	"reflect"
 	"strings"
@@ -15,15 +16,15 @@ import (
 type IndexKeyValueIter interface {
 	// Next returns the next tuple of index key values. The length of the
 	// returned slice will be the same as the number of columns used to
-	// create this iterator.
-	Next() ([]interface{}, error)
+	// create this iterator. The second returned parameter is a repo's location.
+	Next() ([]interface{}, []byte, error)
 	io.Closer
 }
 
 // IndexValueIter is an iterator of index values.
 type IndexValueIter interface {
-	// Next returns the next index value.
-	Next() (interface{}, error)
+	// Next returns the next value (repo's location) - see IndexKeyValueIter.
+	Next() ([]byte, error)
 	io.Closer
 }
 
@@ -31,17 +32,19 @@ type IndexValueIter interface {
 // more functionality by implementing more specific interfaces.
 type Index interface {
 	// Get returns an IndexLookup for the given key in the index.
-	Get(key interface{}) (IndexLookup, error)
+	Get(key ...interface{}) (IndexLookup, error)
 	// Has checks if the given key is present in the index.
-	Has(key interface{}) (bool, error)
+	Has(key ...interface{}) (bool, error)
 	// ID returns the identifier of the index.
 	ID() string
 	// Database returns the database name this index belongs to.
 	Database() string
 	// Table returns the table name this index belongs to.
 	Table() string
-	// Expression returns the indexed expression.
-	Expression() Expression
+	// Expressions returns the indexed expressions. If the result is more than
+	// one expression, it means the index has multiple columns indexed. If it's
+	// just one, it means it may be an expression or a column.
+	ExpressionHashes() []ExpressionHash
 }
 
 // AscendIndex is an index that is sorted in ascending order.
@@ -100,18 +103,20 @@ type Mergeable interface {
 }
 
 // IndexDriver manages the coordination between the indexes and their
-// representation in disk.
+// representation on disk.
 type IndexDriver interface {
 	// ID returns the unique name of the driver.
 	ID() string
-	// Create a new index for the given expression and id.
-	Create(path, db, id string, expr Expression) (Index, error)
-	// Load the index at the given path.
-	Load(path string) (Index, error)
-	// Save the given index at the given path.
-	Save(ctx context.Context, path string, index Index, iter IndexKeyValueIter) error
-	// Delete the index with the given path.
-	Delete(path string, index Index) error
+	// Create a new index. If exprs is more than one expression, it means the
+	// index has multiple columns indexed. If it's just one, it means it may
+	// be an expression or a column.
+	Create(db, table, id string, expressionHashes []ExpressionHash, config map[string]string) (Index, error)
+	// LoadAll loads all indexes for given db and table
+	LoadAll(db, table string) ([]Index, error)
+	// Save the given index
+	Save(ctx context.Context, index Index, iter IndexKeyValueIter) error
+	// Delete the given index.
+	Delete(index Index) error
 }
 
 type indexKey struct {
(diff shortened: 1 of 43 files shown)


kuba--:
### Note
Regarding `Get(key ...interface{}) (IndexLookup, error)` implementation.
Right now, we require `key` parameter to be passed. 
But design doc. says: 
>`Get(key ...interface{}) (IndexLookup, error)`
> ...If no keys are set, valueIterator will go through all > the values (locations) on the index.

If we want to implement _empty_ `Get()` then:
- How we gonna walk through all values (in what order)?
- For multiple expressions what's the expected behaviour
  - Calculate a cartesian with all values for all expressions N x M x ...?
  - Or just walk through all locations for a given index?

sql/index.go
```diff
@@ -111,9 +110,9 @@ type IndexDriver interface {
 	// Create a new index. If exprs is more than one expression, it means the
 	// index has multiple columns indexed. If it's just one, it means it may
 	// be an expression or a column.
-	Create(db, table, id string, expressionHashes []hash.Hash, config map[string]string) (Index, error)
+	Create(db, table, id string, expressionHashes []ExpressionHash, config map[string]string) (Index, error)
 	// Load the index at the given path.
```
ajnavarro: Load -> LoadAll

sql/index/config.go
```diff
@@ -0,0 +1,107 @@
+package index
```
ajnavarro: Should we add some documentation to methods and constructors on this file?

sql/index/pilosa/driver_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	expressions := []sql.ExpressionHash{exh1, exh2}
+
+	d := NewIndexDriver(path)
+	sqlIdx, err := d.Create(db, table, id, expressions, nil)
+	require.Nil(err)
+
+	err = d.Delete(sqlIdx)
+	require.Nil(err)
+}
```
ajnavarro: Can we test what happens if the index generation is stopped abruptly?

sql/index.go
kuba--: Done.

sql/index/config.go
kuba--: Added extra comments. So far I think it's clear what it does.

sql/index/pilosa/driver_test.go
kuba--: We can add some tests for "failing cases", but generally with docker container after restart it will recreate a schema. At the end we won't find anything in bitmaps. The only functions which should work are `Has` and `Get()` (empty get) where we do not touch pilosa.
So what's your opinion?

ajnavarro: Have some tests for failing cases sound great. I'm afraid of having an index loaded that make a query fails just because is corrupted, or an index that appears to work correctly but it has incorrect or partial results.

kuba--:
@ajnavarro - I've added `PilosaHiccup` test, where I restart pilosa container every second. In a meantime I try to `Save/Load/Get` with retry logic.

sql/index/pilosa/driver.go
```diff
@@ -0,0 +1,213 @@
...
+	}, nil
+}
+
+// LoadAll loads all indexes for given db and table
+func (d *Driver) LoadAll(db, table string) ([]sql.Index, error) {
+	root := filepath.Join(d.root, db, table)
+
+	var indexes []sql.Index
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
```
ajnavarro: should we just skip the directory? maybe that error is a permission denied error or something similar.

kuba--: What we can do (except logging)?
- Stop loading all indexes
- Return accumulated error with all dirs which cannot be loaded

sql/index.go
```diff
@@ -285,7 +284,7 @@ func exprListsEqual(a, b []hash.Hash) bool {
 				continue
 			}
 
-			if reflect.DeepEqual(va.Sum(nil), vb.Sum(nil)) {
+			if reflect.DeepEqual(va, vb) {
```
erizocosmico: I think you can just do `va == vb`, since they are byte slices

sql/index/config.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	_, err = w.Write(data)
+	return err
+}
+
+// WriteConfigFile writes the configuration to dir/config.yml file.
+func WriteConfigFile(dir string, cfg *Config) error {
+	path := filepath.Join(dir, ConfigFileName)
```
erizocosmico: this config file is not read and written concurrently right?

sql/index/pilosa/driver.go
erizocosmico: Yep, we should handle that error instead of just skipping it

sql/index/pilosa/driver.go
```diff
@@ -0,0 +1,213 @@
...
+				if err != nil {
+					return err
+				}
+				if !resp.Success {
+					return errors.New(resp.ErrorMessage)
+				}
+			}
+			err = mapping.putLocation(index.Name(), colID, location)
+		}
+	}
```
erizocosmico: what happens if there is an error writing the mapping but everything is already synced with pilosa? Should we implement some kind of _transaction_ for this to ensure data integrity?

sql/index.go
kuba--: Hmm, check: https://play.golang.org/p/_dRDL0lErrV
I get an error: "(slice can only be compared to nil)"

sql/index/config.go
kuba--: should *not* be. Generally it's not thread safe implementation. More like z internal one.

sql/index/pilosa/driver.go
kuba--: Would be great. So far if mapping fails in Save then the function returns error.
We can cal it again and overwrite mapping and pilosa bitmap (what we can find in tests). so at the end should not be a problem just retry.

kuba--:
I've updated `LoadAll` (I accumulate errors) so at the end you can decide do you want to go only with what successfully was loaded or break/stop loading.
Regarding travis I do not know, maybe it's because I check if docker is running before test pilosa.
So I commented these checks as a potential test blockers.

kuba--:
... or maybe just restarting pilosa container in tests hangs a travis randomly (because it works especially when the traffic is low)

kuba--:
I started build manually, we'll see how it goes

kuba--:
When I started manually everything looks fine - weird.

kuba--:
Something is wrong with reporting build status. Build finished with success: https://travis-ci.org/kuba--/go-mysql-server/builds/387725498
and status url says "unknown"

sql/index.go
erizocosmico: Sorry, the ones you can compare are byte **arrays**, not slices haha. Nvm, then

kuba--:
...and build passed

erizocosmico:
As a last small request, can you cleanup a little bit the history of the PR? I see a "Merge" commit in there

erizocosmico:
Shall we merge this @ajnavarro? I'm gonna update gitbase and it'd be nice to get this too.

PR #218, opened 2018-06-14
sql/analyzer: lock indexes per-query and not per-lookup

### LINKED ISSUE #217: Inner join calls RowIter on one of the leafs more than one time, breaking indexes
What `InnerJoin` does is call `RowIter` of the right table once for each row of the left table. Since the node is reused, so is the `sql.IndexLookup`, which cannot be reused because it automatically releases the index when it is done.

erizocosmico:
Fixes #217

### DIFF
--- a/sql/analyzer/analyzer.go
+++ b/sql/analyzer/analyzer.go
@@ -101,7 +101,7 @@ func (a *Analyzer) Analyze(ctx *sql.Context, n sql.Node) (sql.Node, error) {
 		return nil, err
 	}
 
-	for i := 0; !reflect.DeepEqual(prev, cur); {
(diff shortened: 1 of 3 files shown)


erizocosmico:
The real reason is that `reflect.DeepEqual` can't compare `releaser` because of the function it has inside to release. 

From the `reflect.DeepEqual` docs:

> Func values are deeply equal if both are nil; otherwise they are not deeply equal.

So, to make it generic in case any node needs it in the future I added the possibility of implementing an `Equal` method for that. That's an opt-in behavior, only nodes with this special need will need to implement it, otherwise it resorts to good ol' `reflect.DeepEqual`.

PR #219, opened 2018-06-14
sql/index/pilosa: do not error if index directory does not exist

erizocosmico:


sql/index/pilosa/driver.go
```diff
@@ -84,14 +84,18 @@ func (d *Driver) LoadAll(db, table string) ([]sql.Index, error) {
...
+				errors = append(errors, err.Error())
+			}
 			return filepath.SkipDir
 		}
 
 		if info.IsDir() && path != root && info.Name() != "." && info.Name() != ".." {
 			cfg, err := index.ReadConfigFile(path)
 			if err != nil {
-				errors = append(errors, err.Error())
+				if !os.IsNotExist(err) {
```
ajnavarro: this should throw an error because the path exists but no the config file.

erizocosmico:
Fixed @ajnavarro

PR #223, opened 2018-06-15
sql/index: auto delete corrupted indexes

### LINKED ISSUE #220: Auto-detect when pilosa index is corrupt and remove it

erizocosmico:
Closes #220

### DIFF
--- a/sql/index/pilosa/driver.go
+++ b/sql/index/pilosa/driver.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	pilosa "github.com/pilosa/go-pilosa"
+	"github.com/sirupsen/logrus"
 	errors "gopkg.in/src-d/go-errors.v0"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/index"
@@ -62,15 +63,7 @@ func (d *Driver) Create(db, table, id string, expr []sql.ExpressionHash, config 
 		return nil, err
 	}
 
-	return &pilosaIndex{
-		path:        path,
-		client:      d.client,
-		db:          db,
-		table:       table,
-		id:          id,
-		expressions: expr,
-		mapping:     newMapping(path),
-	}, nil
+	return newPilosaIndex(path, d.client, cfg), nil
 }
 
 // LoadAll loads all indexes for given db and table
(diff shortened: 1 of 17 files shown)


sql/index/pilosa/driver.go
```diff
@@ -116,6 +105,50 @@ func (d *Driver) LoadAll(db, table string) ([]sql.Index, error) {
 	return indexes, err
 }
 
+var errCorruptIndex = errors.NewKind("the index in %q is corrupt")
+
+func (d *Driver) loadIndex(path string) (sql.Index, error) {
+	cfg, err := index.ReadConfigFile(path)
```
ajnavarro: Maybe the error is permission denied.

erizocosmico:
Sounds good to me

kuba--:
we can reuse an empty bolt db file. moreover bolt set a lock till is opened

erizocosmico:
@ajnavarro an issue I see with this approach:

 We can know if the index was created or not by checking the `CREATED` file.
But if it's not created, the index is corrupt and we have to delete it. Then we need to read the config, which may fail. What does it happen if reading the config fails? If we don't read the config we can't delete the index because we need that data for removing the pilosa stuff.

After all, it's the same logic as before, adding the extra step of having to check for the `CREATED` file, which is overcomplicating the thing. The only part that is simplified is the part of writing the file at the end of the index save.

I mean, we can use `CREATE`, but we still need to think what to do with config read files and it's basically the same logic with extra files.

Create should not be idempotent because an index cannot be overwritten, only deleted and recreated, right? Or am I missing something?

ajnavarro:
@erizocosmico why do you need to read the config file if the index is corrupt? Just ignore it on LoadAll method, and when someone call again Create, override all the files.

erizocosmico:
So you mean we should leave everything there and wait for recreation? Wouldn't it be better to delete all the garbage that a partial index may have written (which potentially can be a lot)?

kuba--:
So that's why I wanted to reuse boltDB file as a lock. It can give us double-check.
- if file exists
- can you open (if not most likely lock is still set)

erizocosmico:
Changed the logic

sql/index/pilosa/driver.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	pilosa "github.com/pilosa/go-pilosa"
+	"github.com/sirupsen/logrus"
```
kuba--: go-log?

sql/index/pilosa/driver.go
```diff
@@ -116,6 +105,37 @@ func (d *Driver) LoadAll(db, table string) ([]sql.Index, error) {
...
+	}
+
+	if ok {
+		log := logrus.WithFields(logrus.Fields{
+			"err":  err,
+			"path": path,
+		})
+		log.Warn("could not read index file, index is corrupt and will be deleted")
+
+		if err := os.RemoveAll(path); err != nil {
```
kuba--: So when the file exists we just delete it, not checking how long it exists?
it means that we can save and load, over and over

sql/index/pilosa/driver.go
erizocosmico: not yet inside go-mysql-server. Last news i had was to wait until a logging guide came out

sql/index/pilosa/driver.go
erizocosmico: This is only used when the app starts

ajnavarro:
@kuba-- PTAL

PR #225, opened 2018-06-18
sql/expression/function: add function split to the default functions to add to the registry

mcarmonaa:
The `split` function was implemented but not added to the default functions to be registered.

PR #226, opened 2018-06-18
*: added supported sql syntax in new markdown file

mcarmonaa:
I added a new file `SUPPORTED.md` which is linked from the `README.md` to show what is currently supported.

I reviewed by hand what we support right now, so @src-d/data-retrieval have a look please and let me now if I missed something.
 
"Depends" on #225 because the `SPLIT` function is already included as supported here.


SUPPORTED.md
```diff
@@ -0,0 +1,86 @@
...
+- <<
+- \>>
+- &
+- \|
+- ^
+- div
+- %
+
+## Subqueries
+- supported
```
erizocosmico: only as tables, not as expressions

SUPPORTED.md
```diff
@@ -0,0 +1,86 @@
...
+- GROUP BY
+- INSERT INTO
+- LIMIT/OFFSET
+- ORDER BY
+- SELECT
+- SHOW TABLES
+- SORT
+
+## Index expressions
+- CREATE INDEX
```
erizocosmico: I would add that we allow either column names or a single arbitrary expression

SUPPORTED.md
```diff
@@ -0,0 +1,86 @@
...
+- MAX
+- MIN
+
+## Standard expressions
+- ALIAS (AS)
+- LITERAL
+- STAR (*)
+- CAST/CONVERT
+- CREATE TABLE
+- DESCRIBE/DESC/EXPLAIN
```
erizocosmico: I would specify the syntax we use for this (in our implementation `FORMAT=TREE` is required)

kuba--:
LGTM after you apply Miguel's comments.

SUPPORTED.md
```diff
@@ -0,0 +1,86 @@
...
+- MAX
+- MIN
+
+## Standard expressions
+- ALIAS (AS)
+- LITERAL
+- STAR (*)
+- CAST/CONVERT
+- CREATE TABLE
+- DESCRIBE/DESC/EXPLAIN FORMAT=TREE
```
ajnavarro: FORMAT=TREE is only needed on `EXPLAIN [query]`
Maybe we should split this into two commands, `DESCRIBE/DESC/EXPLAIN [tablename]` and `DESCRIBE/DESC/EXPLAIN FORMAT=TREE [query]`

mcarmonaa: done

SUPPORTED.md
```diff
@@ -0,0 +1,87 @@
+# SQL Supported Commands
```
smola: Supported SQL Syntax?

PR #227, opened 2018-06-18
sql/driver/pilosa: implement AscendIndex and DescendIndex

### LINKED ISSUE #211: PilosaDriver: Implement AscendIndex and DescendIndex interfaces
First of all, we need to modify a bit the existing interface to apply all the changes that we did on the base Index interface. 
- Instead of accepting `key interface{}` we need to modify methods to use `key ...interface{}`. 
- `AscendRange(greaterOrEqual, lessThan interface{}) (IndexLookup, error)` becomes `AscendRange(greaterOrEqual, lessThan []interface{}) (IndexLookup, error)`.

The same changes should be applied to DescendIndex interface.

erizocosmico:
Closes #211

### DIFF
--- a/sql/index/pilosa/driver.go
+++ b/sql/index/pilosa/driver.go
@@ -136,7 +136,10 @@ func (d *Driver) loadIndex(path string) (sql.Index, error) {
 	return idx, nil
 }
 
-var errInvalidIndexType = errors.NewKind("expecting a pilosa index, instead got %T")
+var (
+	errInvalidIndexType  = errors.NewKind("expecting a pilosa index, instead got %T")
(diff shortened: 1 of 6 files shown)


sql/index/pilosa/driver.go
```diff
@@ -166,6 +166,10 @@ func (d *Driver) Save(ctx context.Context, i sql.Index, iter sql.IndexKeyValueIt
 		return err
 	}
 
+	// make sure we delete the index in every run before inserting, since there may
+	// be previous data
+	_ = d.client.DeleteIndex(pilosaIndex)
```
mcarmonaa: why isn't it necessary check the error if the index must be deleted?

erizocosmico: Done

PR #232, opened 2018-06-19
sql/(index,plan): add logs for index creation

### LINKED ISSUE #231: Index creation appears to do nothing.
Index creation is a process that runs on background. We should add some log to see some progress.

erizocosmico:
Fixes #231

### DIFF
--- a/sql/plan/create_index.go
+++ b/sql/plan/create_index.go
@@ -116,8 +116,13 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 		return nil, err
 	}
 
+	log := logrus.WithFields(logrus.Fields{
+		"id":     index.ID(),
+		"driver": index.Driver(),
+	})
+
 	go func() {
-		err := driver.Save(ctx, index, iter)
+		err := driver.Save(ctx, index, &loggingKeyValueIter{log: log, iter: iter})
 		close(done)
 		if err != nil {
 			logrus.WithField("err", err).Error("unable to save the index")
@@ -127,8 +132,12 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 			} else {
 				<-deleted
 			}
+		} else {
+			log.Info("index successfully created")
 		}
 	}()
+
+	log.Info("starting to save the index")
 
 	return sql.RowsToRowIter(), nil
 }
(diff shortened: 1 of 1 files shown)


sql/index/pilosa/driver.go
```diff
@@ -183,12 +183,25 @@ func (d *Driver) Save(ctx context.Context, i sql.Index, iter sql.IndexKeyValueIt
...
+
+	var savedRows uint64
 	for colID := uint64(0); err == nil; colID++ {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 
 		default:
+			if savedRows%100 == 0 {
+				log.Debugf("still creating index: %d rows saved so far", savedRows)
```
ajnavarro: Can you add some information about the index id and driver id? something like:

sql/index/pilosa/driver.go
```diff
@@ -220,6 +233,8 @@ func (d *Driver) Save(ctx context.Context, i sql.Index, iter sql.IndexKeyValueIt
 		return err
 	}
 
+	defer log.Debug("finished saving the index")
```
ajnavarro: the same:

sql/index/pilosa/driver.go
erizocosmico: It's a few lines before, it's using `log`, which is defined here: https://github.com/src-d/go-mysql-server/pull/232/files#diff-619845d6f18affa4db50cfe631210577R186

sql/index/pilosa/driver.go
erizocosmico: see prev answer

sql/plan/create_index.go
```diff
@@ -116,6 +116,8 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 		return nil, err
 	}
 
+	log := logrus.WithField("id", index.ID())
+
 	go func() {
 		err := driver.Save(ctx, index, iter)
```
ajnavarro: Instead of adding some custom logs to pilosa driver, we can wrap the IndexKeyValue iterator to have the same logs for all the drivers, WDYT?

erizocosmico: Sounds good. I thought about that, but adding another layer felt a bit too much for just logging, but I don't think the performance loss is gonna be huge here so it makes sense to do it.

erizocosmico:
Done @ajnavarro

PR #234, opened 2018-06-20
sql/(plan,index): force index deletion on save failure

### LINKED ISSUE #233: When index creation fails, deletion throws error because index is not usable

erizocosmico:
Fixes #233 

It was either adding a `force` parameter or a new method that does this. Since the method had to be public I thought I might as well just add the force parameter.


erizocosmico:
Tests fixed, missed a couple usages of the function

erizocosmico:
Can we merge this?

PR #237, opened 2018-06-20
sql/analyzer: fix star resolution when there are natural joins involved

mcarmonaa:
If a natural join isn't replaced yet by the `resolveNaturalJoins` when the `resolveStar` rule is applied, a panic will raise if this natural join is a child of a `Project` or `GroupBy` node.

Related to https://github.com/src-d/gitbase/issues/323


mcarmonaa:
@ajnavarro  done

PR #239, opened 2018-06-21
sql/function/aggregation: add SUM aggregation function

### LINKED ISSUE #236: Add SUM aggregation function

erizocosmico:
Closes #236

### DIFF
--- a/sql/expression/function/aggregation/sum.go
+++ b/sql/expression/function/aggregation/sum.go
@@ -1,0 +1,82 @@
+package aggregation
+
+import (
+	"fmt"
+
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+)
+
+// Sum agregation returns the sum of all values in the selected column.
+// It implements the Aggregation interface.
+type Sum struct {
+	expression.UnaryExpression
+}
+
+// NewSum returns a new Sum node.
+func NewSum(e sql.Expression) *Sum {
+	return &Sum{expression.UnaryExpression{Child: e}}
+}
+
+// Type returns the resultant type of the aggregation.
+func (m *Sum) Type() sql.Type {
+	return sql.Float64
+}
+
+func (m *Sum) String() string {
+	return fmt.Sprintf("SUM(%s)", m.Child)
+}
+
+// TransformUp implements the Transformable interface.
+func (m *Sum) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
+	child, err := m.Child.TransformUp(f)
+	if err != nil {
+		return nil, err
+	}
+	return f(NewSum(child))
+}
+
+// NewBuffer creates a new buffer to compute the result.
+func (m *Sum) NewBuffer() sql.Row {
+	return sql.NewRow(nil)
+}
+
+// Update implements the Aggregation interface.
+func (m *Sum) Update(ctx *sql.Context, buffer, row sql.Row) error {
+	v, err := m.Child.Eval(ctx, row)
+	if err != nil {
+		return err
+	}
+
+	if v == nil {
+		return nil
+	}
+
+	val, err := sql.Float64.Convert(v)
+	if err != nil {
+		val = float64(0)
+	}
+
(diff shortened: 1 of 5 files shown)


sql/expression/function/aggregation/sum.go
```diff
@@ -0,0 +1,144 @@
...
+func (m *Sum) Type() sql.Type {
+	return sql.Int64
+}
+
+func (m *Sum) String() string {
+	return fmt.Sprintf("SUM(%s)", m.Child)
+}
+
+// IsNullable returns whether the return value can be null.
+func (m *Sum) IsNullable() bool {
```
ajnavarro: SUM can be null. If all the values are null, SUM will be null.

sql/expression/function/aggregation/sum.go
```diff
@@ -0,0 +1,144 @@
...
+	expression.UnaryExpression
+}
+
+// NewSum returns a new Sum node.
+func NewSum(e sql.Expression) *Sum {
+	return &Sum{expression.UnaryExpression{Child: e}}
+}
+
+// Resolved implements the Resolvable interface.
+func (m *Sum) Resolved() bool {
```
ajnavarro: We can use UnaryExpression implementation for this method: https://github.com/src-d/go-mysql-server/blob/c065ba6913fb686f7256eac96f08eb87d3b38607/sql/expression/common.go#L18

sql/expression/function/aggregation/sum.go
```diff
@@ -0,0 +1,144 @@
...
+
+		switch b := buffer[0].(type) {
+		case nil:
+			buffer[0] = fv.(float64)
+		case int64:
+			buffer[0] = float64(b) + fv.(float64)
+		case float64:
+			buffer[0] = b + fv.(float64)
+		default:
+			return sql.ErrInvalidType.New(reflect.TypeOf(b))
```
ajnavarro: Sum shouldn't throw an error on unsupported types. `SELECT SUM('test')` returns 0

sql/expression/function/aggregation/sum.go
```diff
@@ -0,0 +1,144 @@
...
+		case int64:
+			buffer[0] = b + iv.(int64)
+		case float64:
+			buffer[0] = b + float64(iv.(int64))
+		default:
+			return sql.ErrInvalidType.New(reflect.TypeOf(b))
+		}
+	}
+
+	spew.Dump(buffer)
```
ajnavarro: leftover?

sql/expression/function/aggregation/sum.go
ajnavarro: When is this error thrown? I check that you have tests for strings too.

erizocosmico:
@ajnavarro So, for `Type` I'm returning `sql.Int64` right now but it can also be `sql.Float64` and there is no way to know that ahead of time. What should we use on those cases?

sql/expression/function/aggregation/sum.go
erizocosmico: yup, every time

sql/expression/function/aggregation/sum.go
erizocosmico: This is for checking the type of the internal buffer. This error should never be returned, it would mean there's a bug in the code. Perhaps it would be better to change it to another error like `this is likely a bug, detected value of type foo in the aggregation buffer`?

erizocosmico:
Changes done @ajnavarro

ajnavarro:
It's a little more tricky than that if we want to emulate MySQL output: https://dev.mysql.com/doc/refman/8.0/en/group-by-functions.html

> For numeric arguments, the variance and standard deviation functions return a DOUBLE value. The SUM() and AVG() functions return a DECIMAL value for exact-value arguments (integer or DECIMAL), and a DOUBLE value for approximate-value arguments (FLOAT or DOUBLE).

erizocosmico:
@ajnavarro but what do we do with the `Type` method? https://github.com/src-d/go-mysql-server/pull/239#issuecomment-399366897

erizocosmico:
Updated to always return float64

Issue #238, opened 2018-06-21
Queries do not work correctly after the index is created


jfontan:
Queries with the indexed expression do not return correct answers. For example:

```
MySQL [(none)]> select file_path, language(file_path, blob_content) from files;
+-----------------------------+-----------------------------------------------+
| file_path                   | language(files.file_path, files.blob_content) |
+-----------------------------+-----------------------------------------------+
| executor.go                 | Go                                            |
| executor_test.go            | Go                                            |
| pack.go                     | Go                                            |
| pack_test.go                | Go                                            |
| server.go                   | Go                                            |
| server_test.go              | Go                                            |
| borges.go                   | Go                                            |
...
| sivafuse/get_path.go        | Go                                            |
| sivafuse/get_path_test.go   | Go                                            |
| sivafuse/git.go             | Go                                            |
| sivafuse/root_siva_fs.go    | Go                                            |
+-----------------------------+-----------------------------------------------+
681 rows in set (0.94 sec)
```

Query with the indexed expression:

```
MySQL [(none)]> select file_path from files where language(file_path, blob_content) = 'Go';
Empty set (0.02 sec)
```

The index was created with:

```
create index language on files using pilosa (language(file_path, blob_content));
```

I've double checked that the index was already created. To be sure that no leftovers were there I've also started with new index directories and started clean pilosa servers.

kuba--:
Root cause:
When we create a pilosa index we call `Create` with one expression:
`language(files.file_path, files.blob_content)` what gives ExpressionHash `83700d6ec9d5a296ba3231e4bd4498786fb9473e`

But then we call `Save` with ` sql.IndexKeyValueIter` which for ` iter.Next()` returns slice of values (file_path, blob_content) instead of one value as a result of `language(files.file_path, files.blob_content)`.

In other words we indexed file paths instead of language names (that's why we cannot find 'Go' or any other language).

For single column indexes everything works fine, e.g.:
```
CREATE INDEX idx_file ON files USING pilosa (file_path);

SELECT * FROM files WHERE file_path = 'LICENSE';
```

kuba--:
Close this one after upgrade mysql

PR #240, opened 2018-06-22
Evaluate expression for index iterator.

kuba--:
@erizocosmico - PTAL if it really makes sense. 
It should fix https://github.com/src-d/go-mysql-server/issues/238, but maybe there is a simpler/nicer fix.

### DIFF
--- a/sql/plan/create_index.go
+++ b/sql/plan/create_index.go
@@ -266,8 +266,9 @@ func getColumnsAndPrepareExpressions(
 }
 
 type evalKeyValueIter struct {
-	iter sql.IndexKeyValueIter
-	eval func(values []interface{}) ([]interface{}, error)
+	ctx   *sql.Context
+	iter  sql.IndexKeyValueIter
+	exprs []sql.Expression
 }
 
 func (eit *evalKeyValueIter) Next() ([]interface{}, []byte, error) {
@@ -275,11 +276,18 @@ func (eit *evalKeyValueIter) Next() ([]interface{}, []byte, error) {
 	if err != nil {
 		return nil, nil, err
 	}
+	row := sql.NewRow(vals...)
+	evals := make([]interface{}, len(eit.exprs))
+	for i, ex := range eit.exprs {
+		eval, err := ex.Eval(eit.ctx, row)
+		if err != nil {
+			return nil, nil, err
+		}
 
-	if eit.eval != nil {
-		vals, err = eit.eval(vals)
+		evals[i] = eval
 	}
-	return vals, loc, err
+
+	return evals, loc, nil
 }
 
 func (eit *evalKeyValueIter) Close() error {
@@ -292,22 +300,7 @@ func getIndexKeyValueIter(ctx *sql.Context, table sql.Indexable, columns []strin
 		return nil, err
 	}
 
-	eit := &evalKeyValueIter{iter: iter}
-	eit.eval = func(values []interface{}) ([]interface{}, error) {
-		row := sql.NewRow(values...)
-		evals := make([]interface{}, len(exprs))
-		for i, ex := range exprs {
-			val, err := ex.Eval(ctx, row)
-			if err != nil {
-				return nil, err
-			}
-
-			evals[i] = val
-		}
-		return evals, nil
-	}
-
-	return eit, err
+	return &evalKeyValueIter{ctx, iter, exprs}, nil
 }
 
 type loggingKeyValueIter struct {
--- a/sql/plan/create_index.go
+++ b/sql/plan/create_index.go
@@ -90,7 +90,7 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 		return nil, ErrInvalidIndexDriver.New(c.Driver)
 	}
 
-	columns, exprs, err := getColumnsAndPrepareExpressions(c.Exprs)
+	columns, exprs, exprHashes, err := getColumnsAndPrepareExpressions(c.Exprs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


sql/plan/create_index.go
```diff
@@ -110,6 +110,10 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 	if err != nil {
 		return nil, err
 	}
+	riter, err := table.RowIter(ctx)
```
erizocosmico: why are we using the `RowIter` and `IndexKeyValueIter`? `IndexKeyValueIter` should already return everything you need to create an index. If there's something missing, it should go there.

The problem is `colNames` is string instead of `sql.Expression`, right? Perhaps then the solution should be to change the `IndexKeyValueIter` signature to have that.

sql/plan/create_index.go
```diff
@@ -275,7 +282,55 @@ func (i *loggingKeyValueIter) Next() ([]interface{}, []byte, error) {
 		i.log.Debugf("still creating index: %d rows saved so far", i.rows)
 	}
 
-	return i.iter.Next()
+	_, loc, err := i.iter.Next()
```
erizocosmico: this feels like too much of a task for `loggingKeyValueIter`, which is just a wrapper for logging

sql/plan/create_index.go
kuba--: yeah, kind of for indexes like *LANG(col1, col2)* `IndexKeyValueIter` returns column names instead of evaluate language. Here I used 2 iterators just like a quick fix/workaround, because  `IndexKeyValueIter`  gives me *location* and with `RowIter` I can `Eval` expression.

sql/plan/create_index.go
kuba--: Yeah, I also had this feeling (especially that I passed a context, what's not good). But this works. Now we need to refine it, so I'm totally open for any suggestion how to make it internally correctly. Maybe now with indexes we should combine these iterators and use one, more generic. For instance sth. what let me `Eval` expressions without `RowIter`.

sql/plan/create_index.go
erizocosmico: But those iterators are not guaranteed to return the rows in the same order, so that's kind of dangerous. I mean, ours might (i'm not really sure), but if someone implements their own most probably not.

sql/plan/create_index.go
erizocosmico: I think probably the best solution would be to change the API of `IndexKeyValueIter` and change `colNames []string` to `expressions []sql.Expression`. Is there any reason we might not wanna do that? @ajnavarro

ajnavarro: colNames should be a list of columns because the table doesn't know about expressions, only columns. So, per example, if we want to index `LANGUAGE(f.file_path,f.blob_content)` the IndexKeyValueIterator should return `file_path` and `blob_content` to be able to execute the `LANGUAGE` Function and index it.

kuba--: And this is what it does right now, but index expects to get _languages_, so somewhere we need an `sql.Expression` and call `Eval` but for this we need a `Row`. At the end we also have to pass a `location`. So this is how it looks so far when I added tons of "glue" to workaround and fix it.
The open question is - how to reshape it internally to get rid of this ugly fix.

kuba--:
If we extend `	IndexKeyValueIter` by _expressions_:
```
IndexKeyValueIter(ctx *Context, colNames []string, expressions []Expression) (IndexKeyValueIter, error)
```

erizocosmico:
@ajnavarro what's the specific reason we need to return the column values instead of the value of the indexed expressions if that's what we're going to store?

ajnavarro:
`IndexKeyValueIter` is used to generate the index, so the implementation must be done by the data source implementor, in this case, `gitbase`. 

`IndexValueIter` is used to retrieve index values from given keys. The implementor of this interface is the specific index, in this case ´pilosa´ index.

I think we shouldn't give the responsibility of evaluating the expression to `gitbase` in this case, and do it in a generic way on `go-mysql-server` side.

To be able to get the expression to evaluate it, first of all, we need to check into the indexes rule if any of the expressions has an index. If yes, we should keep that expression, because we cannot transform an Expression hash into an Expression object.

ajnavarro:
@kuba-- can you rebase instead of creating a merge commit?

kuba--:
I'll send update in a minute

ajnavarro:
@erizocosmico can you have a look please?

sql/plan/create_index.go
```diff
@@ -254,13 +255,59 @@ func getColumnsAndPrepareExpressions(
...
+
+func getIndexKeyValueIter(ctx *sql.Context, table sql.Indexable, columns []string, exprs []sql.Expression) (*evalKeyValueIter, error) {
+	iter, err := table.IndexKeyValueIter(ctx, columns)
+	if err != nil {
+		return nil, err
 	}
 
-	return columns, expressions, nil
+	eit := &evalKeyValueIter{iter: iter}
+	eit.eval = func(values []interface{}) ([]interface{}, error) {
```
erizocosmico: what do you think about passing all needed stuff to the iterator and move this closure to a method in the `evalKeyValueIter` instead?

kuba--: Generally I'm fine with that (basically this is what I previously did in your `loggingKeyValueIter`). The only concerns I had:
- I don't want/like keep `*sqlContext` in structure (it's more to control execution, imo).
- I have the same feeling as you had that adding all `expressions`, `context` and `iterator` makes the structure too heavy.
- Right now creating `&evalKeyValueIter{iter:iter}` (without `eval` implementation) gives you original iterator, what I found maybe useful in testing.

*BUT* it can be just my taste/perception, so if current structure looks complicated/unreadable in your opinion then I'm totally fine to replace _eval closure_ by more fields.

kuba--: Any concerns proofs that it can be potential readability issue . So I've replaced that closure by straight forward, concrete struct and moved logic to `Next`.

PR #258, opened 2018-07-03
sql/*: add some more compatibility with mysqldump

erizocosmico:
This is only a part of what should be implemented to be compatible with mysqldump, more info at: https://github.com/src-d/gitbase/issues/361

What this includes:

- Remove comments.
- Empty queries do not crash (see the /* SOME QUERY INSIDE COMMENT*/).
- `SHOW VARIABLES`, which returns a dummy result.
- `SET`, which returns an dummy result.
- `@@GLOBAL.FOO` and `@@SESSION.FOO` always return `nil` and can be resolved.
- `SHOW DATABASES`.
- `LOCK`,  which returns a dummy result.
- `USE`, which returns a dummy result

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -290,6 +290,30 @@ var queries = []struct {
 			{string("third row3")},
 		},
 	},
+	{
+		`SELECT @@GLOBAL.FOO, @@SESSION.BAR`,
+		[]sql.Row{{nil, nil}},
+	},
+	{
+		`SET @@GLOBAL.FOO = 1`,
+		[]sql.Row{},
+	},
+	{
+		`/*!40101 SET NAMES utf8 */;`,
+		[]sql.Row{},
+	},
+	{
+		`SHOW VARIABLES LIKE 'foo%'`,
+		[]sql.Row{},
+	},
+	{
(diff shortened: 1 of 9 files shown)


ajnavarro:
What do you think about "decouple" all this specific logic to make this compatible with mysqldump?

Maybe we can add some hooks before the query parser to be able to discard or do something with the query string before sending it to the parser.

Regarding specific tables or fields, we can add separated rules for them on the analyzer.

WDYT?

smola:
We don't want to add specific logic for mysqldump. Mysqldump itself is not that important for us, what we would like instead is that every time we add some feature that improves compatibility, other tools benefit. Otherwise, we'll end up with a pile of hacks that only work with mysqldump and Tableau.

erizocosmico:
This does not contain any mysqldump-specific logic. This only contains "hacks" for making some queries not fail. Queries you could do with any kind of MySQL client. They just happen to be some of the ones mysqldump does.

But yes, it's a pile of hacks to make it work (not really, though) with mysqldump. Right solution would be to properly implement all this, but I don't know if we want to do that right now.

PR #260, opened 2018-07-04
sql/analyzer: fix columns in the resolveNaturalJoins rule

mcarmonaa:
Related to https://github.com/src-d/gitbase/issues/365

(When this PR will be merged, a PR updating `go-mysql-server` and adding an integration test in `gitbase`  must be done to close that issue.)

When a natural join is transformed, all the nodes over the transformed node that reference some of the common columns by the right table in the join couldn't resolve them.

With this fix, all those columns resolve now to the left table in the join to be found.

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -449,8 +449,10 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
 		return n, nil
 	}
 
+	var transformed sql.Node
+	var colsToUnresolve = map[string][]string{}
 	a.Log("resolving natural joins, node of type %T", n)
-	return n.TransformUp(func(n sql.Node) (sql.Node, error) {
+	node, err := n.TransformUp(func(n sql.Node) (sql.Node, error) {
 		a.Log("transforming node of type: %T", n)
 		if n.Resolved() {
 			return n, nil
@@ -501,6 +503,7 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
 
 					found = true
 					seen[lcol.Name] = struct{}{}
+					colsToUnresolve[lcol.Name] = []string{lcol.Source, rcol.Source}
 					break
 				}
 			}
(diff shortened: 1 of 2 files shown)


sql/analyzer/rules.go
```diff
@@ -531,14 +534,82 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
...
+		)
+
+		return transformed, nil
+	})
+
+	if err != nil || transformed == nil {
+		return node, err
+	}
+
+	var nodesToUnresolve = []sql.Node{}
```
erizocosmico: s/`var nodesToUnresolve = []sql.Node{}`/`var nodesToUnresolve []sql.Node`/

sql/analyzer/rules.go
```diff
@@ -531,14 +534,82 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
...
+	plan.Inspect(node, func(iNode sql.Node) bool {
```
erizocosmico: why `iNode`?

sql/analyzer/rules.go
```diff
@@ -531,14 +534,82 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
...
+	})
+
+	if len(nodesToUnresolve) == 0 {
+		return node, err
+	}
+
+	return node.TransformUp(func(node sql.Node) (sql.Node, error) {
+		var found bool
+		for _, nu := range nodesToUnresolve {
+			if nu.String() == node.String() {
```
erizocosmico: `reflect.DeepEqual(nu, node)` it's probably less expensive than this and more bullet-proof than relying on the String method of the nodes

sql/analyzer/rules.go
```diff
@@ -531,14 +534,82 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
...
 				join.Right,
 				expression.JoinAnd(conditions...),
 			),
-		), nil
+		)
+
+		return transformed, nil
+	})
+
+	if err != nil || transformed == nil {
```
erizocosmico: can `transformed` be nil in any other case than `err` being `!= nil`?

mcarmonaa: Yes, either there isn't a natural join to transform or it's transformed intto a `cross join` instead of a `inner join`, transformed stays as nil

sql/analyzer/rules.go
mcarmonaa: to avoid use n and node so many times in different context, but yes it looks weird (stands for `inspected node`). Changing it

sql/analyzer/rules.go
```diff
@@ -531,14 +534,82 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
...
+
+		expressioner, ok := node.(sql.Expressioner)
+		if !ok {
+			return node, nil
+		}
+
+		return expressioner.TransformExpressions(func(e sql.Expression) (sql.Expression, error) {
+			var col, table string
+			switch e := e.(type) {
+			case *expression.GetField:
```
kuba--: lets merge these to cases into one:
`
case *expression.GetField, *expression.UnresolvedColumn:
`

mcarmonaa: I can't. `e` would stay as a `sql.Expression` interface so I couldn't access `Name()` and `Table()` methods

PR #265, opened 2018-07-05
sql/analyzer: resolve ascend and descend indexes

### LINKED ISSUE #211: PilosaDriver: Implement AscendIndex and DescendIndex interfaces

### LINKED ISSUE #214: analyzer: Add to resolve index rule range checking using DescendIndex and AscendIndex.
Right now we are checking for equality expressions, we need to implement the same logic but to be able to use range expressions described in DescendIndex and AscendIndex interfaces.

erizocosmico:
Closes #214

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -705,32 +705,102 @@ func TestStarPanic197(t *testing.T) {
 }
 
 func TestIndexes(t *testing.T) {
-	require := require.New(t)
 	e := newEngine(t)
 
 	tmpDir, err := ioutil.TempDir(os.TempDir(), "pilosa-test")
-	require.NoError(err)
+	require.NoError(t, err)
 
-	require.NoError(os.MkdirAll(tmpDir, 0644))
+	require.NoError(t, os.MkdirAll(tmpDir, 0644))
 	e.Catalog.RegisterIndexDriver(pilosa.NewIndexDriver(tmpDir))
 
-	_, _, err = e.Query(sql.NewEmptyContext(), "CREATE INDEX myidx ON mytable (i) WITH (async = false)")
-	require.NoError(err)
+	_, _, err = e.Query(
+		sql.NewEmptyContext(),
+		"CREATE INDEX myidx ON mytable (i) WITH (async = false)",
+	)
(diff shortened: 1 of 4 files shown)


engine_test.go
```diff
@@ -706,48 +706,93 @@ func TestStarPanic197(t *testing.T) {
 }
 
 func TestIndexes(t *testing.T) {
-	require := require.New(t)
```
ajnavarro: why did you remove this?

engine_test.go
```diff
@@ -706,48 +706,93 @@ func TestStarPanic197(t *testing.T) {
...
 	defer func() {
 		done, err := e.Catalog.DeleteIndex("foo", "myidx", true)
-		require.NoError(err)
+		require.NoError(t, err)
 		<-done
 	}()
 
-	_, it, err := e.Query(sql.NewEmptyContext(), "SELECT * FROM mytable WHERE i = 2")
-	require.NoError(err)
+	testCases := []struct {
```
ajnavarro: shall we test indexes made from more than one column too?

ajnavarro: and some AND and OR

engine_test.go
erizocosmico: Because now it's created inside each of the subtests. Otherwise, `require` is redefined and you can't do `require.New` inside

engine_test.go
erizocosmico: We cannot test `AND` and `OR` because there is no `SetOperations` implemented yet in pilosa

erizocosmico: Added the tests for more than one column

ajnavarro:
@erizocosmico rebase and merge please!

PR #268, opened 2018-07-05
 sql/analyzer: split rules in multiple files 

### LINKED ISSUE #267: Split analyzer rules into several files
Should be split as 1 rule per file, unless some rules are related, then they should be in the same file.

erizocosmico:
Closes #267


erizocosmico:
@ajnavarro can we review and merge this? Resolving conflicts in this PR is hell, if we can get this merged before any other thing that touches the analyzer it'd be great.

PR #270, opened 2018-07-06
sql/(plan,expression): add assertions for Unary and Binary nodes and exprs

### LINKED ISSUE #189: Make UnaryNode and BinaryNode interfaces
Right now we have the structs:
```
// UnaryNode is a node that has only one children.
type UnaryNode struct {
	Child sql.Node
}

// BinaryNode is a node with two children.
type BinaryNode struct {
	Left  sql.Node
	Right sql.Node
}
```

They are embedded in the different node implementations but these structs themselves aren't `sql.Node` since they don't implement all methods from a `sql.Node` interface, which is ok  because these two nodes aren't really usable nodes. They just implement those common methods to the specific `sql.Node`s implementations as a fallback for those ones.

Because of this design, there's no way to ask or perform a type assertion on a `sql.Node` to know if that node is a `UnaryNode` or a `BinaryNode`.

I propose make these structs private and expose two interfaces:
```
// UnaryNode is a node that has only one children.
type UnaryNode {
       sql.Node
       IsUnary() bool
}

// BinaryNode is a node with two children.
type BinaryNode {
       sql.Node
       IsBinary() bool
}

type unaryNode struct {
	Child sql.Node
}

type binaryNode struct {
	Left  sql.Node
	Right sql.Node
}
```

So for those nodes embedding `unaryNode` or `binaryNode` they would have to implement the methods `IsBinary() bool {return true}` or `IsUnary() bool {return true}`.

Then, we could identify if a specific implementation of a `sql.Node` is an unary or binary node calling these methods on it. Or if we are handling `sql.Node`, perform type assertions `_, ok := node.(sql.UnaryNode)`.

erizocosmico:
Closes #189


ajnavarro:
Shall we make use of those methods on the code parts that need them?

erizocosmico:
Done. Could only find one usage of that by searching for "unary" and "binary". If we come across any other places in the future we can refactor them

PR #274, opened 2018-07-06
add version udf

smacker:
Implement [VERSION](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_version) udf.

usage:

```go
driver.Catalog.RegisterFunction("version", sql.Function0(function.NewVersion("0.1.1")))
```

```
mysql> select VERSION();
+---------+
| version |
+---------+
| 0.1.1   |
+---------+
1 row in set (0.00 sec)
```

Context: https://github.com/src-d/gitbase-playground/issues/184#issuecomment-402809742

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -16,18 +16,19 @@ type Engine struct {
 }
 
 // New creates a new Engine
-func New(c *sql.Catalog, a *analyzer.Analyzer) *Engine {
+func New(c *sql.Catalog, a *analyzer.Analyzer, versionPostfix string) *Engine {
 	c.RegisterFunctions(function.Defaults)
+	c.RegisterFunction("version", sql.FunctionN(function.NewVersion(versionPostfix)))
+
 	return &Engine{c, a}
 }
 
 // NewDefault creates a new default Engine.
 func NewDefault() *Engine {
 	c := sql.NewCatalog()
-	c.RegisterFunctions(function.Defaults)
-
 	a := analyzer.NewDefault(c)
-	return &Engine{c, a}
+
+	return New(c, a, "go-mysql-server")
 }
 
 // Query executes a query without attaching to any context.
--- a/sql/expression/function/version.go
+++ b/sql/expression/function/version.go
@@ -1,0 +1,49 @@
+package function
+
+import (
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+const mysqlVersion = "8.0.11"
+
+// Version is a function that returns server version.
+type Version string
+
+// NewVersion creates a new Version UDF.
+func NewVersion(versionPostfix string) func(...sql.Expression) (sql.Expression, error) {
+	return func(...sql.Expression) (sql.Expression, error) {
+		// need to return pointer because sql.Expression methods expect pointer receivers
+		e := Version(versionPostfix)
+		return &e, nil
+	}
+}
+
+// Type implements the Expression interface.
+func (f *Version) Type() sql.Type { return sql.Text }
+
+// IsNullable implements the Expression interface.
+func (f *Version) IsNullable() bool {
+	return false
+}
+
+func (f *Version) String() string {
+	return "version"
+}
+
+// TransformUp implements the Expression interface.
+func (f *Version) TransformUp(fn sql.TransformExprFunc) (sql.Expression, error) {
+	return fn(f)
+}
+
+// Resolved implements the Expression interface.
+func (f *Version) Resolved() bool {
+	return true
+}
+
+// Children implements the Expression interface.
+func (f *Version) Children() []sql.Expression { return nil }
+
+// Eval implements the Expression interface.
+func (f *Version) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
+	return mysqlVersion + "-" + string(*f), nil
+}
--- a/engine_test.go
+++ b/engine_test.go
@@ -289,6 +289,12 @@ var queries = []struct {
 			{string("third row3")},
 		},
 	},
+	{
+		"SELECT version()",
+		[]sql.Row{
+			{string("8.0.11-go-mysql-server")},
+		},
+	},
 }
 
 func TestQueries(t *testing.T) {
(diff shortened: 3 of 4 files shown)


ajnavarro:
Sorry @smacker , but we are trying to stick the new functionality as much as possible to MySQL behaviors.

The way to implement that is using `SHOW VARIABLES` statement, that is not implemented right now:

```
SHOW VARIABLES LIKE "%version%";

+-------------------------+----------------------------+
| Variable_name           | Value                      |
+-------------------------+----------------------------+
| innodb_version          | 5.6.36-82.2                |
| protocol_version        | 10                         |
| slave_type_conversions  |                            |
| version                 | 10.1.29-MariaDB            |
| version_comment         | MariaDB Server             |
| version_compile_machine | x86_64                     |
| version_compile_os      | Linux                      |
| version_malloc_library  | system                     |
| version_ssl_library     | OpenSSL 1.1.0g  2 Nov 2017 |
| wsrep_patch_version     | wsrep_25.21                |
+-------------------------+----------------------------+
10 rows in set (0.02 sec)
```

WDYT @smola ?

smola:
I wonder if third party tools expect `VERSION()` to conform to some regular expression to check for compatibility with specific MySQL major versions.

The `SHOW VARIABLES` approach, with `gitbase_version` and a `go_mysql_server_version` versions look safer in principle.

smola:
@ajnavarro Looking at [VERSION](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_version) docs, a build-dependent suffix is expected. So we could use something like `8.0.11-gitbase0.2.3`.

ajnavarro:
So, we must return the MySQL version that we are "trying to emulate", right?

sql/functionregistry.go
```diff
@@ -21,6 +21,8 @@ type Function interface {
 }
 
 type (
+	// Function0 is a function with 0 arguments.
+	Function0 func() Expression
```
ajnavarro: Shall we use FunctionN() with no arguments (And ignore the unexpected arguments) for that function to avoid more code?

smola:
@ajnavarro Yep. Since we're not trying to emulate any specific older version, we could just use the highest version as of today, such as 8.0.11.

sql/expression/function/version.go
```diff
@@ -0,0 +1,47 @@
+package function
+
+import (
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// Version is a function that returns server version.
+type Version struct {
```
kuba--: if that's it then maybe we can go with `type Version string`

sql/functionregistry.go
kuba--: either just `Function(...)` or `Function0()` + `FunctionN(...)`

smacker:
👍 thanks! `SHOW VARIABLES` looks better indeed!

erizocosmico:
Only problem with `SHOW VARIABLES` is that we have no such thing as a mechanism to store and retrieve session/global variables yet, so it's a bit of a harder effort to do than this.

ajnavarro:
@smacker that is a good approach, but as @smola said, we need to return the standard MySQL format for that function: `mysqlVersion-prefix`. So in our case will be: `8.0.11-gitbase[VERSION]`

erizocosmico:
Perhaps the part of the mysql version should be hidden from the user and `NewVersion("foo")` should be `WHATEVERMYSQLVERSION-foo` internally (notice gitbase is not there, as this is go-mysql-server, that is gitbases's job to add it)?

smacker:
- change type of the udf to string
- Removed Function0, use FunctionN and ignore arguments
- I didn't understand @kuba-- comment about `either just Function(...) or Function0() + FunctionN(...)`
    * Version can’t implement `Function` interface because of private `sql.isFunction()` method (version lives in different package)
    * And I don’t get `Function0() + FunctionN(...)` at all

This repository doesn't know anything about gitbase, does it?

@ajnavarro 

> You must add the function into the function registry default functions: 
I don't see how I can do it. This udf needs input version from a constructor.

The current proposal in engine.go.
By default, it returns `8.0.11-go-mysql-server`, but an application can override it using non-default engine constructor.

engine.go
```diff
@@ -8,25 +8,29 @@ import (
...
 type Engine struct {
 	Catalog  *sql.Catalog
 	Analyzer *analyzer.Analyzer
 }
 
 // New creates a new Engine
-func New(c *sql.Catalog, a *analyzer.Analyzer) *Engine {
+func New(c *sql.Catalog, a *analyzer.Analyzer, versionPostfix string) *Engine {
 	c.RegisterFunctions(function.Defaults)
+	c.RegisterFunction("version", sql.FunctionN(function.NewVersion(mysqlVersion+"-"+versionPostfix)))
```
smola: `mysqlVersion` should probably always be used, an only postfix be given to `NewVersion`. Or make `mysqlVersion` public, since its `go-mysql-version` the one that implements MySQL stuff and controls supported features.

smacker: Could you please help me to understand which part of `go-mysql-version` defines mysql version?

Originally my reasoning was `engine` controls supported features. So I could create `my-engine` and add more features from newer version. That's why I put constant here.
Do you think it should be defined in `Version` UDF itself? Or any other place?

kuba--:
I meant just one function or 2 for zero and at least one parameter
so if we cannot have just `Function(...)`, I would stick with 2 functions `Function0()` and `FunctionN(...)`

engine.go
```diff
@@ -8,25 +8,29 @@ import (
...
+
 	return &Engine{c, a}
 }
 
 // NewDefault creates a new default Engine.
 func NewDefault() *Engine {
 	c := sql.NewCatalog()
-	c.RegisterFunctions(function.Defaults)
```
ajnavarro: why did you remove this?

smacker: because now `NewDefault` calls `New` constructor which registers function.

Before `NewDefault` was returning `Engine` instance so copy-paste in both constructors was required.

engine.go
ajnavarro: `mysqlVersion` should be inside the `Version` function. So at the end you only need to send to the constructor `versionPostfix`:

ajnavarro:
Just a small change: https://github.com/src-d/go-mysql-server/pull/274#discussion_r201624675 , rebase and LGTM

smacker:
- moved mysql version to udf
- rebased on master
- squashed

PR #275, opened 2018-07-06
sql/plan: allow synchronous index creations

### LINKED ISSUE #271: Sync create index queries
There are some cases, such as tests or gitbase regression tests where you don't want to create the indexes in the background because you need them to be created before you start doing queries.

Do this manually requires a lot of knowledge about the insides and they are susceptible to change. By optionally letting the user decide if they should be sync or async we can abstract all this away and just use a query to create them.

It could be some config or environment variable.

WDYT? @ajnavarro

erizocosmico:
Closes #271

### DIFF
--- a/sql/index.go
+++ b/sql/index.go
@@ -467,9 +467,12 @@ func exprListsEqual(a, b []ExpressionHash) bool {
 // expression or id while the other is still being created.
 // When something is sent through the returned channel, it means the index has
 // finished it's creation and will be marked as ready.
-func (r *IndexRegistry) AddIndex(idx Index) (chan<- struct{}, error) {
+// Another channel is returned to notify the user when the index is ready.
+func (r *IndexRegistry) AddIndex(
+	idx Index,
+) (created chan<- struct{}, ready <-chan struct{}, err error) {
 	if err := r.validateIndexToAdd(idx); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	r.mut.Lock()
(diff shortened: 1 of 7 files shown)


sql/plan/create_index.go
```diff
@@ -43,12 +44,16 @@ func NewCreateIndex(
 	driver string,
 	config map[string]string,
 ) *CreateIndex {
+	async := config["async"] != "false"
+	// Remove async from the config, since it's not meant for the driver.
+	delete(config, "async")
```
ajnavarro: It doesn't do any bad keeping the property there.

sql/plan/create_index.go
```diff
@@ -124,22 +129,32 @@ func (c *CreateIndex) RowIter(ctx *sql.Context) (sql.RowIter, error) {
...
+
+	log.WithField("async", c.Async).Info("starting to save the index")
 
-	log.Info("starting to save the index")
+	if c.Async {
+		go createIndex()
+	} else {
+		createIndex()
+		// wait for the index status to be updated
+		time.Sleep(50 * time.Millisecond)
```
ajnavarro: We should find a better solution for this.

erizocosmico:
Fixed

sql/plan/create_index.go
```diff
@@ -49,6 +50,7 @@ func NewCreateIndex(
 		Exprs:  exprs,
 		Driver: driver,
 		Config: config,
+		Async:  config["async"] != "false",
```
mcarmonaa: Is it ok using the configuration keys directly as strings instead of parameterize them through constants?

sql/index.go
```diff
@@ -479,15 +482,17 @@ func (r *IndexRegistry) AddIndex(idx Index) (chan<- struct{}, error) {
 	r.indexOrder = append(r.indexOrder, key)
 	r.mut.Unlock()
 
-	var created = make(chan struct{})
+	var _created = make(chan struct{})
```
mcarmonaa: Just curious about the underscore prefix, Is it frequently used or kind of standard in the community?

erizocosmico: Nope, `created` and `ready` were used as named return arguments in the signature for documentation purposes, but they are restricted to one direction, so I can't really use them inside

sql/plan/create_index.go
erizocosmico: I don't think async is gonna be called anything else anytime soon, and it's not used anywhere else, so no need to add two more constants for this

erizocosmico:
Shall we merge this?

Issue #212, opened 2018-06-13
PilosaDriver: Implement SetOperations and Mergeable interfaces on IndexLookup


ajnavarro:
To allow the use of several indexes of the same type, we need to implement this interfaces on the actual PilosaDriver index implementation.

PR #281, opened 2018-07-09
Implementation of Mergeable and SetOperations for pilosa's lookups

kuba--:
! Should be merged after: https://github.com/src-d/go-mysql-server/pull/277 !

This PR. Implements 2 interfaces (Mergeable and SetOperations) for following lookups:
- indexLookup
- ascendLookup
- descendLookup

It closes: https://github.com/src-d/go-mysql-server/issues/212

### simplified example

```      
       unionLookup := lookup0.Union(lookup1, lookup2)

       ascLookup := pilosaIdx.AscendLessThan(keys...) 
       diffLookup := ascLookup.Difference(unionLookup)
```


kuba--:
Ok, right now is  squashed and rebased (+ some fixes)

engine.go
```diff
@@ -16,19 +16,18 @@ type Engine struct {
 }
 
 // New creates a new Engine
-func New(c *sql.Catalog, a *analyzer.Analyzer, versionPostfix string) *Engine {
+func New(c *sql.Catalog, a *analyzer.Analyzer) *Engine {
```
ajnavarro: that code is not well rebased, you are removing changes from master.

kuba--:
Ok, now it looks it contains only my chnages.

sql/index/pilosa/driver_test.go
```diff
@@ -349,6 +350,364 @@ func TestAscendDescendIndex(t *testing.T) {
...
+
+	interLookup, ok = lookupPath.(sql.SetOperations)
+	require.True(ok)
+	interIt, err = interLookup.Intersection(lookupLang).Values()
+	require.NoError(err)
+	loc, err := interIt.Next()
+	require.NoError(err)
+	require.True(bytes.Equal(loc, itPath.records[0].location))
+	_, err = interIt.Next()
+	require.True(err == io.EOF)
```
ajnavarro: maybe you can use require.Equal in that cases.

kuba--: I afraid that it won't work for byte slices, but looks that it works, so I've updated.

PR #283, opened 2018-07-10
sql/analyzer: add rule to evaluate filters ahead of time

### LINKED ISSUE #280: Create optimization rule to evaluate filter expressions if possible
We should add an optimization rule to evaluate beforehand filter expressions, if possible. Per example, in this case:
```
SELECT * FROM commits WHERE 1=0
```

we can just return 0 results without asking the data source.

erizocosmico:
Closes #280

This commit adds the `eval_filters` rule to the analyzer, which evaluates all evaluable filters in `Filter` nodes and removes any unnecessary filter. If it turns out the final result of evaluating the filters is true, independently of the non-evaluable expressions, the filter node is just removed. If it's false, the filter node is removed and replaced with a special node called `Nothing` that returns an empty result set while still holding the instance of the inner node.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -295,6 +295,10 @@ var queries = []struct {
 			{string("8.0.11-go-mysql-server")},
 		},
 	},
+	{
+		"SELECT * FROM mytable WHERE 1 > 5",
+		[]sql.Row{},
+	},
 }
 
 func TestQueries(t *testing.T) {
--- a/sql/analyzer/optimization_rules.go
+++ b/sql/analyzer/optimization_rules.go
@@ -295,3 +295,108 @@ func expressionSources(expr sql.Expression) []string {
 
 	return result
 }
+
+func evalFilter(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
+	if !node.Resolved() {
+		return node, nil
+	}
+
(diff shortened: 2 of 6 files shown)


ajnavarro:
I would change a bit the logic here to do not use a special `Nothing` filter.

Maybe instead of using a special filter implementation, we can modify the query tree to remove the filter and all the nodes at the bottom for a special node called `EmptyTable`

You can see the SparkSQL rule here: https://github.com/apache/spark/blob/d54d8b86301581142293341af25fd78b3278a2e8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala#L806-L832

This can work nicely with another rule that is in charge to merge Filters (but this can be done in another PR):

https://github.com/apache/spark/blob/d54d8b86301581142293341af25fd78b3278a2e8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/Optimizer.scala#L747-L760

ajnavarro:
@erizocosmico Sounds good.

erizocosmico:
Done @ajnavarro

erizocosmico:
I can add it, but since integration tests on engine only check results, you don't know if it's because it was replaced by `EmptyTable` or because it ran the query and it returned no results.

erizocosmico:
Added @ajnavarro

PR #286, opened 2018-07-12
sql/analyzer: add rule to reorder aggregations inside projections

erizocosmico:
Closes https://github.com/src-d/gitbase/issues/378

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -299,6 +299,14 @@ var queries = []struct {
 		"SELECT * FROM mytable WHERE 1 > 5",
 		[]sql.Row{},
 	},
+	{
+		"SELECT SUM(i) + 1, i FROM mytable GROUP BY i ORDER BY i",
+		[]sql.Row{
+			{float64(2), int64(1)},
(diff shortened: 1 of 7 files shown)


ajnavarro:
Just a question, that rule is there because other rule or a set of rules are returning an invalid query tree?

If yes, instead of do another rule, we need to fix the main problem.

erizocosmico:
I will make it generate a correct tree. But the problem will still be there after the tree is generated correctly, which what this rule solves (just like `reorder_projection` does with projections).

erizocosmico:
Fixed the original issue, which ended up being a parse issue (just top level expression was being checked to be an expression).

This allowed the removal of some code in the rule, since there can't be aggregations in projects now, only GroupBy's.

Also, noticed a case I wasn't handling here, when there are several aggregations inside the same expression, which now has a test for it and is working correctly.

The need for the rule is that a top level expression that's not an Aggregation does not have `Update`, `Merge` and so on, so it's a bit hard to evaluate. But now the trees are, at least, correct after the parse, which were not before.

PR #287, opened 2018-07-12
sqle: change Engine constructor to accept a Config struct

### LINKED ISSUE #285: Introduce better config mechanism to engine constructor
After adding the version UDF we introduced a new argument `versionPostfix` to the `sqle.New` function. Tomorrow, it may be something else.

We should replace that with either a struct that can be later on given more fields or functional arguments (functional arguments, for example, wouldn't have broken the build of gitbase as API would be the same).

erizocosmico:
Closes #285


engine.go
```diff
@@ -9,14 +9,28 @@ import (
...
 	Catalog  *sql.Catalog
 	Analyzer *analyzer.Analyzer
 }
 
-// New creates a new Engine
-func New(c *sql.Catalog, a *analyzer.Analyzer, versionPostfix string) *Engine {
+// New creates a new Engine with custom configuration. To create an Engine with
+// the default settings use `NewDefault`.
+func New(c *sql.Catalog, a *analyzer.Analyzer, cfg *Config) *Engine {
+	var versionPostfix = "go-mysql-server"
```
ajnavarro: shall we just do not register the "VERSION()" function if that config property is empty? Or maybe just return the MySQL version and that's it

erizocosmico: That way, `NewDefault` will always return just the MySQL version

erizocosmico:
Done @ajnavarro

ajnavarro:
Fix broken test and LGTM.

erizocosmico:
Should be fixed now

PR #296, opened 2018-07-19
sql/analyzer: fix resolveNaturalJoins rule to work properly with table aliases

mcarmonaa:
Partial fix for https://github.com/src-d/gitbase/issues/388

Related to https://github.com/src-d/gitbase/pull/391

### DIFF
--- a/sql/analyzer/resolve_natural_joins.go
+++ b/sql/analyzer/resolve_natural_joins.go
@@ -8,6 +8,11 @@ import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
 )
 
+type transformedJoin struct {
+	node     sql.Node
+	condCols map[string][]string
+}
+
 func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 	span, ctx := ctx.Span("resolve_natural_joins")
 	defer span.Finish()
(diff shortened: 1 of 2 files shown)


sql/analyzer/resolve_natural_joins.go
```diff
@@ -179,3 +176,68 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
 		})
 	})
 }
+
+func isOverTranformedNode(node sql.Node, transformed []*transformedJoin) (over bool, colsToUnresolve map[string][]string) {
+	node.TransformUp(func(n sql.Node) (sql.Node, error) {
```
erizocosmico: if you're not using the result, you should be using `plan.Inspect` instead (with which you can probably use `==` instead of `reflect.DeepEqual`

mcarmonaa:
`qualifyColumns` works fine. The bug comes from the `NATURAL JOIN`, the columns involved in the condition after changed it to a `INNER JOIN` must be referenced by the left table.

If this change (included the table alias) is not handle in the `resolveNaturalJoins` rule, then in the `resolveColumns` rule a well qualified column will be found but referenced by the table in the right of the join, so actually that column will not be found because in the child's schema at these point the correct source for the column is the table in the left of the join.

PR #300, opened 2018-07-24
sql/analyzer: fix resolveNaturalJoins rule for chained natural joins

mcarmonaa:
Related to https://github.com/src-d/gitbase/issues/388 and https://github.com/src-d/gitbase/pull/391.

Fix to handle properly column resolution for chained natural joins.

### DIFF
--- a/sql/analyzer/resolve_natural_joins.go
+++ b/sql/analyzer/resolve_natural_joins.go
@@ -10,7 +10,12 @@ import (
 
 type transformedJoin struct {
 	node     sql.Node
-	condCols map[string][]string
+	condCols map[string]*transformedSource
+}
+
+type transformedSource struct {
+	correct string
+	wrong   []string
 }
 
 func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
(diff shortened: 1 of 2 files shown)


sql/analyzer/resolve_natural_joins.go
```diff
@@ -135,9 +149,14 @@ func resolveNaturalJoins(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, e
 		return node, err
 	}
 
+	var transformedSeen bool
 	return node.TransformUp(func(node sql.Node) (sql.Node, error) {
-		above, colsToUnresolve := isOverTranformedNode(node, transformed)
-		if !above {
+		if ok, _ := isTransformedNode(node, transformed); ok {
```
kuba--: here I think we can simplify by:
```
transformedSeen, _ = isTransformedNode(node, transformed)
return node, nil
```

mcarmonaa: I can't do that. `transformedSeen` is a variable outside the scope of the function executed for each node by the `node.TransformUp` method.

If I always assign the `transformedSeen` variable, it will be false the next time a non transformed node is evaluated although a previous node is a transformed node.

I'm not sure if I explained myself well or if I misunderstood you @kuba--

kuba--: Ok, I know what you mean - nvm.

PR #314, opened 2018-08-01
sql: change Index and IndexDriver interfaces and remove ExpressionHash logic

### LINKED ISSUE #305: Move `ExpressionHash` logic to the driver
We have this logic exposed from the `sql/core.go` level:

```go
// ExpressionHash is a SHA-1 checksum
type ExpressionHash []byte

// NewExpressionHash returns a new SHA1 hash for given Expression instance.
// SHA1 checksum will be calculated based on ex.String().
func NewExpressionHash(ex Expression) ExpressionHash {
	h := sha1.Sum([]byte(ex.String()))
	return ExpressionHash(h[:])
}

// DecodeExpressionHash  decodes a hexadecimal string to ExpressionHash
func DecodeExpressionHash(hexstr string) (ExpressionHash, error) {
	h, err := hex.DecodeString(hexstr)
	if err != nil {
		return nil, err
	}
	return ExpressionHash(h), nil
}

// EncodeExpressionHash encodes an ExpressionHash to hexadecimal string
func EncodeExpressionHash(h ExpressionHash) string {
	return hex.EncodeToString(h)
}
```

And we are offering a way to deal with `ExpressionHash`es directly from the `Index` interface through the method `ExpressionHashes`:

```go
// Index is the basic representation of an index. It can be extended with
// more functionality by implementing more specific interfaces.
type Index interface {
	// Get returns an IndexLookup for the given key in the index.
	Get(key ...interface{}) (IndexLookup, error)
	// Has checks if the given key is present in the index.
	Has(key ...interface{}) (bool, error)
	// ID returns the identifier of the index.
	ID() string
	// Database returns the database name this index belongs to.
	Database() string
	// Table returns the table name this index belongs to.
	Table() string
	// Expressions returns the indexed expressions. If the result is more than
	// one expression, it means the index has multiple columns indexed. If it's
	// just one, it means it may be an expression or a column.
	ExpressionHashes() []ExpressionHash
	// Driver ID of the index.
	Driver() string
}
```

The thing is that all this is a specific detail of the pilosa driver implementation so we can move all this logic to there and don't exposed it as part of the API.

Instead, we can have a method `Expressions` which returns the expressions indexed by an index as the string representation of an `Expression`:

```go
// Index is the basic representation of an index. It can be extended with
// more functionality by implementing more specific interfaces.
type Index interface {
	// Get returns an IndexLookup for the given key in the index.
	Get(key ...interface{}) (IndexLookup, error)
	// Has checks if the given key is present in the index.
	Has(key ...interface{}) (bool, error)
	// ID returns the identifier of the index.
	ID() string
	// Database returns the database name this index belongs to.
	Database() string
	// Table returns the table name this index belongs to.
	Table() string
    // Expressions returns the indexed expressions as the string representation of a expression.
    //If the result is more than one expression, it means the // index has multiple columns indexed. If it's just one, it
    // means it may be an expression or a column.
	Expressions() []string
	// Driver ID of the index.
	Driver() string
}
```

`IndexDriver` interface will need to change its API:
```go
NOW
---
// IndexDriver manages the coordination between the indexes and their
// representation on disk.
type IndexDriver interface {
	// ID returns the unique name of the driver.
	ID() string
	// Create a new index. If exprs is more than one expression, it means the
	// index has multiple columns indexed. If it's just one, it means it may
	// be an expression or a column.
	Create(db, table, id string, expressionHashes []ExpressionHash, config map[string]string) (Index, error)
	// LoadAll loads all indexes for given db and table
	LoadAll(db, table string) ([]Index, error)
	// Save the given index
	Save(ctx *Context, index Index, iter IndexKeyValueIter) error
	// Delete the given index.
	Delete(index Index) error
}

AFTER (Create method signature changed)
-----
// IndexDriver manages the coordination between the indexes and their
// representation on disk.
type IndexDriver interface {
	// ID returns the unique name of the driver.
	ID() string
	// Create a new index. If exprs is more than one expression, it means the
	// index has multiple columns indexed. If it's just one, it means it may
	// be an expression or a column.
	Create(db, table, id string, expressions []string, config map[string]string) (Index, error)
	// LoadAll loads all indexes for given db and table
	LoadAll(db, table string) ([]Index, error)
	// Save the given index
	Save(ctx *Context, index Index, iter IndexKeyValueIter) error
	// Delete the given index.
	Delete(index Index) error
}

------
(MAYBE
    Create(db, table, id string, expressions []sql.Expression, config map[string]string) (Index, error)

AND RELY ON THE DRIVER TO DO THE STRING CONVERSION?)
------
```

Doing this we can get all the expressions indexed. One of the case of use for this is the `show indexes` implementation. For the moment is not showing all the needed information because there's no way to get all the indexed expressions.

If we need build an `Expression` instance I guess we could parse the string. Anyway, parsing this is not something we need for the moment. Outside the driver implementation we are using `ExpressionHash` to compare expressions.

`ExpressionHash` is the hash of the string representation of an `Expression` so we can compare directly the strings.

To be able to use `Expressions() []string` from the current pilosa index/driver implementation, we need to store these expressions as plain text in the index's `config.yml` file.

Right now we are storing the expressions' hashes in the config file, but if we store the expressions' strings too, maybe it's not necessary store the hashes.

We can calculate the hashes from the string on the fly when creating or loading an index. So the specific pilosa index implementation will keep the hashes internally for its own purposes and will keep the string expressions to expose them through the `Index` interface methods.

WDYT @src-d/data-retrieval ?

mcarmonaa:
Closes #305 

- Changed `Index` and `IndexDriver` interfaces to remove the use of `ExpressionHash`
- Removed all the logic related to `ExpressionHash`.
- Now the `config.yml` file for pilosa indexes doesn't store expression hashes anymore:
```yml
db: ""
table: refs
id: refs_name_idx
expressions:
- refs.ref_name
drivers:
  pilosa: {}
```

### DIFF
--- a/sql/index/pilosa/index.go
+++ b/sql/index/pilosa/index.go
@@ -1,21 +1,12 @@
 package pilosa
 
 import (
-	"crypto/sha1"
-
 	errors "gopkg.in/src-d/go-errors.v1"
 
 	pilosa "github.com/pilosa/go-pilosa"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/index"
 )
-
-type expressionHash []byte
-
-func newExpressionHash(ex string) expressionHash {
-	h := sha1.Sum([]byte(ex))
-	return expressionHash(h[:])
-}
 
 // pilosaIndex is an pilosa implementation of sql.Index interface
 type pilosaIndex struct {
@@ -27,15 +18,9 @@ type pilosaIndex struct {
 	table       string
 	id          string
 	expressions []string
-	hashes      []expressionHash
 }
 
 func newPilosaIndex(path string, client *pilosa.Client, cfg *index.Config) *pilosaIndex {
-	hashes := make([]expressionHash, len(cfg.Expressions))
-	for i, e := range cfg.Expressions {
-		hashes[i] = newExpressionHash(e)
-	}
-
 	return &pilosaIndex{
 		path:        path,
 		client:      client,
@@ -43,7 +28,6 @@ func newPilosaIndex(path string, client *pilosa.Client, cfg *index.Config) *pilo
 		table:       cfg.Table,
 		id:          cfg.ID,
 		expressions: cfg.Expressions,
-		hashes:      hashes,
 		mapping:     newMapping(path),
 	}
 }
@@ -76,7 +60,7 @@ func (idx *pilosaIndex) Get(keys ...interface{}) (sql.IndexLookup, error) {
 		index:       index,
 		mapping:     idx.mapping,
 		keys:        keys,
-		expressions: idx.expressionHashes(),
+		expressions: idx.expressions,
 	}, nil
 }
 
@@ -92,7 +76,7 @@ func (idx *pilosaIndex) Has(key ...interface{}) (bool, error) {
 
 	// We can make this loop parallel, but does it make sense?
 	// For how many (maximum) keys will be asked by one function call?
-	for i, expr := range idx.expressionHashes() {
+	for i, expr := range idx.expressions {
 		name := frameName(idx.ID(), expr)
 
 		val, err := idx.mapping.get(name, key[i])
@@ -126,11 +110,6 @@ func (idx *pilosaIndex) Expressions() []string {
 	return idx.expressions
 }
 
-// expressionHashes returns the hashes of the indexed expressions.
-func (idx *pilosaIndex) expressionHashes() []expressionHash {
-	return idx.hashes
-}
-
 func (pilosaIndex) Driver() string { return DriverID }
 
 func (idx *pilosaIndex) AscendGreaterOrEqual(keys ...interface{}) (sql.IndexLookup, error) {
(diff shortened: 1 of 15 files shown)


ajnavarro:
the expression `refs.ref_name` will work if in one query we are using just `ref_name` ?

erizocosmico:
@ajnavarro All columns are always qualified during analysis. I'm not sure if aliases are removed, though. Can you check, @mcarmonaa?

sql/index/pilosa/index.go
```diff
@@ -1,13 +1,22 @@
...
+	"crypto/sha1"
+
 	errors "gopkg.in/src-d/go-errors.v1"
 
 	pilosa "github.com/pilosa/go-pilosa"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/index"
 )
 
+type expressionHash []byte
```
kuba--: I think, we totally don't need it. We can totally really on expressions as strings

mcarmonaa:
@kuba-- I had thought the hashes were needed for index and frame names but now I see there is a hash calculation anyway to get those names prefixing the expressions.

should I remove `ExpressionHash` then?

kuba--:
@mcarmonaa yeah, this is what I meant. We either way calculate it after we moved all table indexes into the one pilosa index. So I would say - go ahead and remove it.

mcarmonaa:
`ExpressonHash` related logic has been totally removed

Issue #311, opened 2018-08-01
Partitions: Change Indexable interfaces to make use of partitions


ajnavarro:
Base IndexLookup interface will change to be able to get IndexValueIters per Partition.

```go

type IndexLookup interface {
	Values(Partition) (IndexValueIter, error)
}

type IndexableTable interface {
    Table
    WithIndexLookup(IndexLookup) Table
    IndexKeyValues(Context, []string, Partition) IndexKeyValueIter
}

```

erizocosmico:
This is partly done in #329 

Only missing thing is making index drivers work with partitions

erizocosmico:
Since now we have partitions, we should store the partitions in different pilosa frames. Right now, we still send an iterator with everything to the index to create it, so we don't know which rows belong to each partition.

I think we should change the `Save` method of the `IndexDriver` interface to accept a `IndexableTable` instead of an `IndexKeyValueIter`. That way, the driver can iterate partitions and get the `IndexKeyValueIter`s for each partition and save them in different frames.

WDYT @ajnavarro?

kuba--:
Question - shall we add partitions to the driver index (and/or mapping) to quickly get everything  just from particular partition?

ajnavarro:
@erizocosmico  yep, totally needed.

erizocosmico:
@ajnavarro just updated with a second possibility, which one you think it's better?

erizocosmico:
@kuba-- yep

erizocosmico:
I think for now we can get by with not storing info about partitions in the index or mapping. Gitbase can get the partitions from the repository pool and the mem table has them hardcoded, so it's fine for now.

Issue #310, opened 2018-08-01
Data source api: change Project and Filter pushdown interfaces


ajnavarro:
The interfaces will be:

```go
type FilteredTable interface {
	Table
	HandledFilters(filters []Expression) []Expression
	WithFilters(filters []Expression) Table
}

type ProjectedTable interface {
	Table
	WithProjection(colNames []string) Table
}
```

Issue #309, opened 2018-08-01
Partitions: Make tables use partitions


ajnavarro:
Change Table interface to:

```go
type Partition interface {
	Key() []byte
}

type PartitionIter interface {
	io.Closer
	Next() (Partition, error)
}

type Table interface {
	Schema() Schema
      Partitions(ctx Context) PartitionIter
      PartitionRows(Partition) (Rowiter, error)
}
```

Now, Table is not implementing Node interface. A new ResolvedTable Node will get Tables from the catalog and use them to iterate Rows.

PR #329, opened 2018-08-10
*: partitions interfaces and ResolvedTable node

mcarmonaa:
- Interfaces from #309 #310 #311 has been added.
- A new node `plan.ResolvedTable`has been created.
- `mem.Table` now implements the interfaces from #309 #310 #311
- Now some rules are excuted only once before and after the rest of the `DefaultRules`
- `CreateIndex` now iters over all the table's partitions.
- The pushdown rules now make use of the interfaces from #310 #311 and the `plan.ResolvedTable`, so old nodes from `plan/pushdown.go` have been removed.
- Some rules has been fixed to works with the new interfaces.
- All the test have needed changes to use the `mem.Table`, `plan.ResolvedTable` and the new interfaces.

Note that the specific implementation for the lookups haven't changed and don't comply with the new `IndexLookup` interface to handle partitions, so the code under `sql/index/pilosa` and `sql/index/pilosalib` doesn't compile.

I commented two test in `engine_test.go` that made use of the pilosa indexes. When the pilosa indexes implement the `IndexLookup` interfaces, they should work.

Everything else seems to work properly. :+1:


sql/plan/resolved_table.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"fmt"
+	"io"
+
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// ResolvedTable represents a resolved SQL Table.
+type ResolvedTable struct {
```
kuba--: Does it mean that for `ResolvedTable` we keep also `Node` interface?

sql/plan/resolved_table.go
```diff
@@ -0,0 +1,110 @@
...
+
+	row, err := i.rows.Next()
+	if err != nil && err == io.EOF {
+		i.partition = nil
+		if err := i.rows.Close(); err != nil {
+			return nil, err
+		}
+
+		i.rows = nil
+		return i.Next()
```
kuba--: tail recursion is fine, I just thought about `goto` ;)

sql/plan/resolved_table.go
```diff
@@ -0,0 +1,110 @@
...
+	table      sql.TableV2
+	partitions sql.PartitionIter
+	partition  sql.Partition
+	rows       sql.RowIter
+}
+
+func (i *tableIter) Next() (sql.Row, error) {
+	if i.partition == nil {
+		partition, err := i.partitions.Next()
+		if err != nil {
```
kuba--: if it's EOF we don't need to close it?

sql/plan/resolved_table.go
mcarmonaa: Is this a case where is not considered a "bad practice" use `goto`?

sql/plan/resolved_table.go
kuba--: No, I think you're right resolved table needs a Node, but in pure table we have to get rid of node.

sql/plan/resolved_table.go
kuba--: No, do not use `goto`. Tail recursion is fast and we won't have huge amount of partitions (I suppose). Personally I just like `goto` - it's fast ;)

sql/plan/resolved_table.go
mcarmonaa: yes, you can see a dummy implementation of  `sql.TableV2` in the test used by the `ResolvedTable` node.

ajnavarro:
@mcarmonaa what is needed to merge this?

sql/analyzer/pushdown.go
```diff
@@ -9,6 +9,27 @@ import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
 )
 
+func eraseProjection(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
```
erizocosmico: why is this on pushdown? it's an optimization rule

mem/table.go
```diff
@@ -2,86 +2,279 @@ package mem
...
 
+var _ sql.Table = (*Table)(nil)
+var _ sql.Inserter = (*Table)(nil)
+var _ sql.FilteredTable = (*Table)(nil)
+var _ sql.ProjectedTable = (*Table)(nil)
+var _ sql.IndexableTable = (*Table)(nil)
+
 // NewTable creates a new Table with the given name and schema.
-func NewTable(name string, schema sql.Schema) *Table {
+func NewTable(name string, schema sql.Schema, numPartitions int) *Table {
```
erizocosmico: since most of the calls are with 0 numPartitions, how about having `NewTable` and `NewTablePartitioned` for when you want partitions?

mem/table.go
```diff
@@ -2,86 +2,279 @@ package mem
...
 import (
 	"bytes"
-	"encoding/binary"
+	"encoding/gob"
 	"fmt"
 	"io"
+	"strconv"
 
-	errors "gopkg.in/src-d/go-errors.v1"
+	errors "gopkg.in/src-d/go-errors.v0"
```
ajnavarro: why did you change to previous go-errors version?

erizocosmico: that was probably go-imports' thing

mcarmonaa: It was indeed

erizocosmico:
Tests are failing btw

erizocosmico:
@mcarmonaa can you rebase against `partitions`? I'm working on the exchange rule based on this branch and I need to get the exchange implementation too.

erizocosmico:
Just rebased the PR, adapted the exchange and fixed an issue with inserter. `Insert` method was inside the `Table` interface, which is not how it should be.

It still fails to compile both index driver packages, since they don't support partitions yet.

erizocosmico:
@kuba-- yeah, it's going against a feature branch, so it's not a problem that it does not compile. It's not trivial to make it work with partitions, since it requires a lot of changes. I'm doing that in another PR that uses this as its base.

engine_test.go
```diff
@@ -770,157 +764,157 @@ func TestStarPanic197(t *testing.T) {
...
-	defer func() {
-		time.Sleep(1 * time.Second)
-		done, err := e.Catalog.DeleteIndex("foo", "myidx", true)
-		require.NoError(err)
-		<-done
-
-		require.NoError(os.RemoveAll(tmpDir))
-	}()
-}
+// func TestIndexes(t *testing.T) {
```
ajnavarro: shall we uncomment this code, even if it is failing? just to do not forget about that.

mem/database.go
```diff
@@ -41,6 +41,5 @@ func (d *Database) Create(name string, schema sql.Schema) error {
 	}
 
 	d.tables[name] = NewTable(name, schema)
-
```
ajnavarro: unnecessary modified file.

mem/table.go
```diff
@@ -2,86 +2,284 @@ package mem
...
 		}
 	}
 
-	t.data = append(t.data, row.Copy())
 	return nil
 }
 
-func (t Table) String() string {
+// String implements the sql.Table inteface.
+func (t *Table) String() string {
```
ajnavarro: This is still needed at this level or can be implemented at ResolvedTable level?

sql/plan/resolved_table.go
```diff
@@ -0,0 +1,106 @@
...
+	sql.Table
+	name string
+}
+
+var _ sql.Table = (*ResolvedTable)(nil)
+var _ sql.Node = (*ResolvedTable)(nil)
+var _ sql.Nameable = (*ResolvedTable)(nil)
+
+// NewResolvedTable creates a new instance of ResolvedTable.
+func NewResolvedTable(name string, table sql.Table) *ResolvedTable {
```
ajnavarro: If table has name, why do we need to set a name on resolvedTable too?

mem/table.go
mcarmonaa: it can be implemented at ResolvedTable level, but the table implementation can provide a better information since there's no way to know if there are filters, projections or indexes set in the table from the ResolvedTable

sql/plan/resolved_table.go
mcarmonaa: `mem.Table` has name, the interface `sql.Table` hasn't it.

erizocosmico: Maybe `sql.Table` should have a name

engine_test.go
erizocosmico: Don't worry, I am fixing those tests in my PR. If he uncomments that code, the root package will not compile

PR #330, opened 2018-08-14
sql/plan: implement Exchange node

### LINKED ISSUE #312: Partitions: Exchange operator.
An Exchange will be a Node in charge to parallelize part of the query tree.
The parallelization level will be provided as Analyzer instantiation property.
The Exchange node will basically execute the same part of the tree behind it with a different partition.

Exchange nodes will call to Partitionable.Partitions() method to get the Partition iterator.

It will send that partition to the nethermost node, that must be a Table.
The number of partitions executed at the same time depends on the parallelization provided to the Analyzer.

Exchange node will return EOF when PartitionIter returns EOF and all the executed node threads return EOF.

erizocosmico:
Closes #312 

**Note:** Since #309 is not implemented yet, I added some basic interfaces to be able to implement this. When #309 is implemented, this will need a minor refactor to adapt to the shape of the final interfaces.

### DIFF
--- a/sql/plan/exchange.go
+++ b/sql/plan/exchange.go
@@ -1,0 +1,309 @@
+package plan
+
+import (
+	"fmt"
+	"io"
+	"sync"
+
+	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// ErrNoPartitionable is returned when no Partitionable node is found
+// in the Exchange tree.
+var ErrNoPartitionable = errors.NewKind("no partitionable node found in exchange tree")
+
+// Exchange is a node that can parallelize the underlying tree iterating
+// partitions concurrently.
+type Exchange struct {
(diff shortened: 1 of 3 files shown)


sql/plan/exchange.go
```diff
@@ -0,0 +1,309 @@
...
+		Parallelism: parallelism,
+	}
+}
+
+// RowIter implements the sql.Node interface.
+func (e *Exchange) RowIter(ctx *sql.Context) (sql.RowIter, error) {
+	var partitionable sql.Partitionable
+	Inspect(e.Child, func(n sql.Node) bool {
+		if p, ok := n.(sql.Partitionable); ok {
+			partitionable = p
```
ajnavarro: should we check that there is only one partitionable?

erizocosmico: I guess the rule will already do that, right?

ajnavarro:
LGTM when you wish @erizocosmico

PR #336, opened 2018-08-21
 sql/analyzer: add rule to parallelize adding exchanges 

### LINKED ISSUE #313: Partitions: Rules to apply Exchange operator into query trees
The following actions should be implemented as analyzer rules to be able to parallelize queries:
- If parallelization is > 1, we can introduce an Exchange node, if not, the rule is doing nothing.

- Exchange nodes should be added on top of much nodes as possible.

- We stop adding nodes inside Exchange node when we found a binary node or the last node with no parents. We must also stop adding nodes inside Exchange node if we find an Aggregation. In the future, we can parallelize that kind of nodes too.

- We cannot add Exchange nodes on top of other Exchange nodes, but we can have several Exchange nodes on siblings (see example graph).

- Exchange nodes must finish always in a table.

erizocosmico:
Depends on #329 

Closes #313


ajnavarro:
@erizocosmico could you rebase when you have time please?

erizocosmico:
Rebased. But the rebase is not gonna fix the CI. That will happen when #311 is done

ajnavarro:
@erizocosmico merge when you are ready

PR #345, opened 2018-09-05
sql/analyzer: don't parallelize index operations

erizocosmico:


ajnavarro:
@erizocosmico check Travis. Appears that changing ` a.Parallelism <= 0` to ` a.Parallelism <= 1 ` is causing some errors on tests.

PR #347, opened 2018-09-05
fix: convert nil value properly for blob type

mcarmonaa:
Before this fix, when a nil value was returned for a column with type BLOB, it did panic.


mcarmonaa:
@ajnavarro  done

PR #352, opened 2018-09-06
*: implement show full processlist

### LINKED ISSUE #348: Show full process list
See: https://github.com/src-d/gitbase/issues/454#issuecomment-418718875

erizocosmico:
Closes #348

### DIFF
--- a/server/context.go
+++ b/server/context.go
@@ -10,7 +10,7 @@ import (
 	"gopkg.in/src-d/go-vitess.v0/mysql"
 )
 
-// SessionBuilder creates sessions given a context and a MySQL connection.
+// SessionBuilder creates sessions given a MySQL connection and a server address.
 type SessionBuilder func(conn *mysql.Conn, addr string) sql.Session
 
 // DoneFunc is a function that must be executed when the session is used and
@@ -35,7 +35,7 @@ type SessionManager struct {
 	contexts        map[uuid.UUID]context.CancelFunc
 }
 
-// NewSessionManager creates a SessionManager with the given ContextBuilder.
+// NewSessionManager creates a SessionManager with the given SessionBuilder.
 func NewSessionManager(
 	builder SessionBuilder,
 	tracer opentracing.Tracer,
--- a/server/context.go
+++ b/server/context.go
@@ -11,21 +11,22 @@ import (
 )
 
 // SessionBuilder creates sessions given a context and a MySQL connection.
-type SessionBuilder func(*mysql.Conn) sql.Session
+type SessionBuilder func(conn *mysql.Conn, addr string) sql.Session
 
 // DoneFunc is a function that must be executed when the session is used and
 // it can be disposed.
 type DoneFunc func()
 
 // DefaultSessionBuilder is a SessionBuilder that returns a base session.
-func DefaultSessionBuilder(_ *mysql.Conn) sql.Session {
-	return sql.NewBaseSession()
+func DefaultSessionBuilder(c *mysql.Conn, addr string) sql.Session {
+	return sql.NewSession(addr, c.User)
 }
 
 // SessionManager is in charge of creating new sessions for the given
 // connections and keep track of which sessions are in each connection, so
 // they can be cancelled is the connection is closed.
 type SessionManager struct {
+	addr            string
 	tracer          opentracing.Tracer
 	mu              *sync.Mutex
 	builder         SessionBuilder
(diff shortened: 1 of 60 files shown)


sql/analyzer/process.go
```diff
@@ -0,0 +1,174 @@
...
+				return n, nil
+			}
+
+			t, err := wrapTable(n.Name(), n.Table)
+			if err != nil {
+				return nil, err
+			}
+
+			return plan.NewResolvedTable(n.Name(), t), nil
+		case sql.Table:
```
ajnavarro: is this necessary? Why we cannot just get ResolvedTables?

erizocosmico: Not all tables may be inside ResolvedTables (e.g. SquashedTables)

sql/analyzer/pushdown_test.go
```diff
@@ -228,5 +228,5 @@ func TestPushdownIndexable(t *testing.T) {
 	})
 	require.NoError(err)
 
-	require.Equal(expected, result)
+	require.Equal(expected, removeProcessNodes(t, result))
```
ajnavarro: I would prefer to add process node to expected or remove the rule to the actual analyzer. WDYT?

erizocosmico: It can't be added to the expected, because the nodes have functions and they're not comparable, so I guess I can remove it from the analyzer

sql/analyzer/process.go
ajnavarro: So, if I understood correctly, maybe the easiest solution is to move the table name to the sql.Table interface, and apply the rule just to `sql.Table`s, right?

erizocosmico: Yup. Not having the name in the Table is a PITA a lot of times, and there's no reason a table can't know its name (after all, it needs to return it as the `Source` in the `Schema` method, so it must know).

erizocosmico:
@ajnavarro Table now implements sql.Nameable again, removed the name param from ResolvedTable (because it's already on the table) and removed the rule from the analyzer instead of the result nodes

erizocosmico:
Also rebased

sql/analyzer/process.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+
+			return plan.NewResolvedTable(t), nil
+		case sql.Table:
+			t, err := wrapTable(n)
+			if err != nil {
+				return nil, err
+			}
+
+			return plan.NewResolvedTable(t), nil
```
ajnavarro: if the table is inside a resolved table, we are adding two resolved table nodes?

sql/analyzer/process.go
```diff
@@ -0,0 +1,166 @@
...
+
+		seen[name] = struct{}{}
+		return &processTable{t, func() {
+			processList.UpdateProgress(pid, name, 1)
+		}}, nil
+	}
+
+	n, err := n.TransformUp(func(n sql.Node) (sql.Node, error) {
+		switch n := n.(type) {
+		case *plan.ResolvedTable:
```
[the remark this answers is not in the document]
erizocosmico: If we don't keep this, we would wrap `ResolvedTable`s in `ResolvedTable`s in the next `case`

sql/analyzer/process.go
erizocosmico: That's exactly what the previous case is for

sql/analyzer/process.go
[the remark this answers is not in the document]
ajnavarro: then we can remove 
```
t, err := wrapTable(n.Table)
if err != nil {
	return nil, err
}
return plan.NewResolvedTable(t), nil
```
Because all the tables in resovled tables will be already wrapped, right?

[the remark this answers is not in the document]
erizocosmico: Nope, `sql.Table` is not a node and is not reached by `TransformUp`, only via `ResolvedTable`. The second case is for `sql.Table`s that are also nodes.

server/context.go
```diff
@@ -11,21 +11,22 @@ import (
 )
 
 // SessionBuilder creates sessions given a context and a MySQL connection.
-type SessionBuilder func(*mysql.Conn) sql.Session
+type SessionBuilder func(conn *mysql.Conn, addr string) sql.Session
```
mcarmonaa: The comment above is outdated.

server/context.go
```diff
@@ -35,8 +36,13 @@ type SessionManager struct {
 }
 
 // NewSessionManager creates a SessionManager with the given ContextBuilder.
-func NewSessionManager(builder SessionBuilder, tracer opentracing.Tracer) *SessionManager {
+func NewSessionManager(
```
mcarmonaa: The comment above is outdated.

erizocosmico:
@mcarmonaa comments fixed

@ajnavarro removed the case for `sql.Table`. Just so this can be implemented in gitbase, I also changed the schema source validation rule to just check that is not empty.

PR #363, opened 2018-09-14
pilosalib: fix index usage with parallelism

erizocosmico:
This solves https://github.com/src-d/gitbase/issues/472


sql/index/pilosalib/mapping.go
```diff
@@ -35,21 +35,35 @@ func newMapping(path string) *mapping {
...
 
 // openCreate opens and sets creation mode in the database.
-func (m *mapping) openCreate(create bool) {
+func (m *mapping) openCreate(create bool) error {
 	m.clientMut.Lock()
 	defer m.clientMut.Unlock()
+	m.mut.Lock()
+	defer m.mut.Unlock()
+
+	if m.clients == 0 && m.db == nil {
```
kuba--: so maybe instead of having here 2 mutexes we can make `m.clients` a `sync.atomic` var. WDYT?

erizocosmico: that other mutex it's protecting `created` as well

PR #368, opened 2018-09-19
sql/analyzer: resolve aggregate aliases in groupby grouping

erizocosmico:
Closes https://github.com/src-d/gitbase/issues/460

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -128,6 +128,14 @@ var queries = []struct {
 		},
 	},
 	{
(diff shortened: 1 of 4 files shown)


erizocosmico:
@kuba-- it's explained in the comment before those lines. The columns are stored in a map, if we didn't sort them, the output would not be idempotent. Note that even if the columns of the projection are sorted, the output of the query does not change because there's no change in the order of the groupby aggregate.

PR #372, opened 2018-09-20
*: correctly handle process management using processlist

### LINKED ISSUE #371: Fix behaviour of query cancellation
Continuation of #369 

- Drop connection does not kill any query
- Kill query kills the query but does not drop the connection
- Kill drops the connection after killing all executing queries

This affects index creation and so on.

erizocosmico:
Closes #371 

Summary of changes:

- Session manager, no longer tracks contexts created for queries, only sessions. Contexts are managed by the ProcessList as part of the Process.
- Pid is available from Context now, and connection id from the Session.
- trackProcess rule does no longer register the process, only attaches step information and wraps the node if needed to track completion.
- KILL correctly handled: query just kills query but keeps connection, connection drops connection and all associated queries.
- KILL connection is also tested now (using unsafe, though :( )
- DROP INDEX cannot drop indexes that are currently being created.

### DIFF
--- a/server/context.go
+++ b/server/context.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 
 	opentracing "github.com/opentracing/opentracing-go"
-	uuid "github.com/satori/go.uuid"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-vitess.v0/mysql"
 )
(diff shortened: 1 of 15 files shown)


sql/analyzer/process.go
```diff
@@ -64,5 +55,11 @@ func trackProcess(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 		return nil, err
 	}
 
-	return plan.NewQueryProcess(n, func() { processList.Done(pid) }), nil
+	// Don't wrap CreateIndex in a QueryProcess, as it is a CreateIndexProcess.
+	// CreateIndex will take care of marking the process as done on its own.
+	if _, ok := n.(*plan.CreateIndex); ok {
```
ajnavarro: why we cannot do the same as standard queries with create index queries? they are processing partitions too, and the process is measuring processed partitions.

server/context.go
```diff
@@ -60,43 +63,24 @@ func (s *SessionManager) NewSession(conn *mysql.Conn) {
...
-	s.mu.Unlock()
-
-	return context, func() {
-		s.mu.Lock()
-		defer s.mu.Unlock()
+	context := sql.NewContext(
+		context.Background(),
+		sql.WithSession(sess),
+		sql.WithTracer(s.tracer),
+		sql.WithPid(s.nextPid()),
```
kuba--: maybe pid can be atomic - wdyt?
to avoid in the future situations that `nextPid()` will block because mutex wasn't unlocked.
People are used to write:
```
lock()
defer unlock()
```
if we accidentally do sth. like that here, we will stuck in `nextPid`

erizocosmico: I mean, it could be atomic. But it's a given that we will unlock all mutexes at all times 😛 
We already use the mutex for other things, anyway, so the mutex will still be used even if it's not for this.

sql/analyzer/process.go
ajnavarro: @erizocosmico ping

erizocosmico: This has nothing to do with measuring. If we wrap it with a QueryProcess the process will be marked as done as soon as the query returns, which is not when it has finished in the case of async indexes.

PR #373, opened 2018-09-20
sql/index: remove pilosa driver and rename pilosalib to pilosa

### LINKED ISSUE #370: Remove pilosa index driver
We already have a serverless implementation using pilosa, `pilosalib`. We should remove old `pilosa` and rename `pilosalib` to `pilosa`. Also we should take that into account on documentation.

erizocosmico:
Closes #370


jfontan:
As @kuba-- it would be nice to have a note somewhere about the change. It may be enough in the release notes as this library is not directly user facing. Still, we have to keep in mind this change on gitbase and maybe provide the migration path described.

PR #377, opened 2018-09-24
sql/analyzer: execute parallelize as last rule

erizocosmico:
Fixes https://github.com/src-d/gitbase/pull/486

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -223,14 +223,6 @@ var queries = []struct {
 		},
 	},
 	{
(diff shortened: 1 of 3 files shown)


engine_test.go
```diff
@@ -331,6 +323,33 @@ func TestQueries(t *testing.T) {
 		}
 	})
 }
+func TestDescribe(t *testing.T) {
```
erizocosmico: in parallel it has different output, instead of adding two cases for all other queries, i preferred to extract this one to its own test

PR #380, opened 2018-09-24
sql/analyzer: also check node if it is not wrapped in query process

erizocosmico:


erizocosmico:
@kuba-- when have we used `may`? I personally find `should` much more clear than `may`

kuba--:
I saw it in Manu's PRS. Maybe is more optional. So it's up to you

PR #382, opened 2018-09-25
scripts: go-vitess v1

mcuadros:
Update of go-vitess to the last upstream


ajnavarro:
I'm having the following errors trying to use vitess.v1, so maybe we should improve that script: 

```
sql/type.go:15:2: code in directory /home/antonio/work/src/gopkg.in/src-d/go-vitess.v1/vt/proto/query expects import "vitess.io/vitess/go/vt/proto/query"
```

mcuadros:
Should work know, vitess, renamed the org so was replacing the old org instead of the new one.

Also, the changed the log pattern, so now is used glog, we should find a clean way to replace it with `go-log`.

kuba--:
@mcuadros - I had to remove unused packages to compile it.
PTAL: https://github.com/src-d/go-vitess/pull/3
This one works for me.

kuba--:
I've simplified (for our needs) a Makefile, so maybe we can use this one:
https://github.com/src-d/go-mysql-server/blob/ef2dfbb889f5eeff92209871af19269ebeee5c12/_scripts/go-vitess/Makefile

PR #386, opened 2018-09-28
sql: make table resolution case insensitive

### LINKED ISSUE #385: Table names must be case insensitive
To avoid incompatibility problems, table names should be case-insensitive. `TABLENAME` and `tablename` must resolve to the same table.

erizocosmico:
Fixes #385

### DIFF
--- a/sql/catalog.go
+++ b/sql/catalog.go
@@ -1,6 +1,8 @@
 package sql
 
 import (
+	"strings"
+
 	"gopkg.in/src-d/go-errors.v1"
 )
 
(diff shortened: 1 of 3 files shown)


sql/catalog.go
```diff
@@ -51,9 +53,20 @@ func (d Databases) Table(dbName string, tableName string) (Table, error) {
...
+
 	tables := db.Tables()
-	table, found := tables[tableName]
-	if !found {
+	// Try to get the table by key, but if the name is not the same,
+	// then use the slow path and iterate over all tables comparing
+	// the name.
+	table, ok := tables[tableName]
+	if !ok {
+		for name, table := range tables {
```
kuba--: I thought that we'll store table names only lowercase  (so on `Create` table we'll convert everything to lowercase). It will avoid extra _looping_ every time.

erizocosmico: We do, other implementations of `Database` may not

PR #396, opened 2018-10-03
add integration tests for a variety of 3rd party clients

erizocosmico:
Don't merge. The CI stuff can only be tested on CI so it may require some changes before working.

Adds tests for:

- pymysql
- python mysql
- mysql ruby (the one written in ruby, included in rails)
- go
- dotnet mysqlconnector
- mariadb for java
- nodejs
- php

Basically, it starts the example server and does a simple query. Because what we're checking here is that those clients are able to connect and query, that is enough.

All the tests follow the same structure to make it easier to develop new tests.

```
/_integration
    /<test>
        Makefile
```

All tests must include a `Makefile` with the `test` task. Setting up dependencies and so on is the responsibility of each test.

Then, there is a `/_integration/run <test>` script that launches the example server, runs the test and kills the example server.

For ease of use, they can be launched with `make TEST=<test> integration` from the root of the project.


erizocosmico:
Tests are passing now! :tada:

PR #401, opened 2018-10-04
sql: expression: function: Add CEIL, CEILING, FLOOR, ROUND. [Fixes #393]

### LINKED ISSUE #393: Add support for ROUND, FLOOR, CEIL
* https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_ceil
* https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_round
* https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_floor

Zarkopafilis:
Signed-off-by: Theodoros Zarkopafilis Ntakouris <zarkopafilis@gmail.com>

### DIFF
--- a/sql/expression/function/ceil_round_floor.go
+++ b/sql/expression/function/ceil_round_floor.go
@@ -2,10 +2,13 @@ package function
 
 import (
 	"fmt"
-	"gopkg.in/src-d/go-mysql-server.v0/sql"
-	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 	"math"
 	"reflect"
+)
+
+import (
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 )
 
 // Ceil returns the smallest integer value not less than X.
@@ -53,7 +56,7 @@ func (c *Ceil) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 	}
 
 	if !sql.IsNumber(c.Child.Type()) {
-		return 0, nil
+		return int32(0), nil
 	}
 
 	if !sql.IsDecimal(c.Child.Type()) {
@@ -115,7 +118,7 @@ func (f *Floor) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 	}
 
 	if !sql.IsNumber(f.Child.Type()) {
-		return 0, nil
+		return int32(0), nil
 	}
 
 	if !sql.IsDecimal(f.Child.Type()) {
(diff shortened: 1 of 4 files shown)


Zarkopafilis:
#393

sql/expression/function/ceil_round_floor.go
```diff
@@ -0,0 +1,258 @@
...
+func (r *Round) String() string {
+	if r.Right == nil {
+		return fmt.Sprintf("ROUND(%s, 0)", r.Right.String())
+	}
+
+	return fmt.Sprintf("ROUND(%s, %s)", r.Right.String(), r.Left.String())
+}
+
+// Resolved implements the Expression interface.
+func (r *Round) Resolved() bool {
```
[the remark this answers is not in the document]
Zarkopafilis: I disagree with this. ROUND() shouldn't break when Right is not specified (nil)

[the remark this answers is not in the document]
erizocosmico: Yes, ROUND should not break when Right is not specified. That's why the constructor has to be `NewRound(args ...sql.Expression) (sql.Expression, error)` or two arguments will always be required. You can see an example of a function with optional arguments here: https://github.com/src-d/go-mysql-server/blob/master/sql/expression/function/substring.go#L23-L38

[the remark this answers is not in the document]
Zarkopafilis: @erizocosmico So if right is not specified (nil) it doesn't mean that we should let the default functions of binary expression handle Children, IsNullable and Resolved. It needs special care when right is nil (don't return it). If it **evaluates** to nil it's a different case, that causes no problem here.

sql/expression/function/ceil_round_floor.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if child == nil {
+		return nil, nil
+	}
+
+	if !sql.IsNumber(c.Child.Type()) {
+		return 0, nil
```
erizocosmico: It should be `int32(0)`

sql/expression/function/ceil_round_floor.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if child == nil {
+		return nil, nil
+	}
+
+	if !sql.IsNumber(f.Child.Type()) {
+		return 0, nil
```
erizocosmico: it should be `int32(0)`

sql/expression/function/ceil_round_floor.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// NewRound returns a new Round expression.
+func NewRound(args ...sql.Expression) (sql.Expression, error) {
+	argLen := len(args)
+	if argLen == 0 || argLen > 2 {
+		return nil, sql.ErrInvalidArgumentNumber.New("1 or 2", argLen)
+	}
+
+	d := func(i int) sql.Expression {
```
erizocosmico: What do you think about this for avoiding that extra closure?

```go
var right sql.Expression = nil
if len(args) == 2 {
        right = args[1]
}
```

sql/expression/function/ceil_round_floor.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if xVal == nil {
+		return nil, nil
+	}
+
+	if !sql.IsNumber(r.Left.Type()) {
+		return 0, nil
```
erizocosmico: Should be `int32(0)`

sql/expression/function/ceil_round_floor.go
Zarkopafilis: In favor of performance, yes.

(I am too influenced by other languages nowdays..)

PR #419, opened 2018-10-05
sql: expression: function: Add LOWER, UPPER [Fix #390]

### LINKED ISSUE #390: Add support for LOWER, UPPER
Add support for LOWER and UPPER:
* https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_lower
* https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_upper

Zarkopafilis:
#390 

Signed-off-by: Zarkopafilis <zarkopafilis@gmail.com>

### DIFF
--- a/sql/expression/function/lower_upper.go
+++ b/sql/expression/function/lower_upper.go
@@ -1,0 +1,120 @@
+package function
+
+import (
+	"fmt"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+	"strings"
+)
+
+// Lower is a function that returns the lowercase of the text provided.
+type Lower struct {
+	expression.UnaryExpression
+}
+
+// NewLower creates a new Lower expression.
+func NewLower(e sql.Expression) sql.Expression {
+	return &Lower{expression.UnaryExpression{Child: e}}
+}
+
+// Eval implements the Expression interface.
+func (l *Lower) Eval(
+	ctx *sql.Context,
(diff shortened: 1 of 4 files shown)


sql/expression/function/lower_upper.go
```diff
@@ -0,0 +1,120 @@
...
+	v, err := l.Child.Eval(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+
+	if v == nil {
+		return nil, nil
+	}
+
+	if !sql.IsText(l.Child.Type()) {
```
erizocosmico: According to MySQL behaviour, the argument is converted to text, then lowered. So you can just do:

```go
v, err = sql.Text.Convert(v)
if err != nil {
        return nil, err
}

return strings.ToLower(v.(string)), nil
```

sql/expression/function/lower_upper.go
```diff
@@ -0,0 +1,120 @@
...
+	v, err := u.Child.Eval(ctx, row)
+	if err != nil {
+		return nil, err
+	}
+
+	if v == nil {
+		return nil, nil
+	}
+
+	if !sql.IsText(u.Child.Type()) {
```
erizocosmico: According to MySQL behaviour, the argument is converted to text, then lowered. So you can just do:

```go
v, err = sql.Text.Convert(v)
if err != nil {
        return nil, err
}

return strings.ToUpper(v.(string)), nil
```

ajnavarro:
@Zarkopafilis can you rebase and resolve master conflicts please?

PR #427, opened 2018-10-08
sql: implement LIKE expression

erizocosmico:
Closes https://github.com/src-d/go-mysql-server/issues/426

### DIFF
--- a/sql/expression/like.go
+++ b/sql/expression/like.go
@@ -1,0 +1,135 @@
+package expression
+
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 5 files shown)


sql/expression/like.go
```diff
@@ -0,0 +1,135 @@
+package expression
+
+import (
+	"bytes"
+	"fmt"
+	"regexp"
```
ajnavarro: can you use `https://github.com/src-d/go-mysql-server/tree/master/internal/regex` ?

erizocosmico:
Fixed and rebased @ajnavarro

PR #448, opened 2018-10-11
*: implement LOCK and UNLOCK of tables

### LINKED ISSUE #410: Implement table locking
Implement `LOCK` and `UNLOCK` for tables. This should prevent tables from being written / updated while they're locked.

See: https://github.com/src-d/gitbase/issues/361#issuecomment-427285397

erizocosmico:
Closes #410

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1361,6 +1361,42 @@ func TestUse(t *testing.T) {
 	require.Equal("foo", e.Catalog.CurrentDatabase())
 }
 
+func TestLocks(t *testing.T) {
+	require := require.New(t)
+
+	t1 := newLockableTable(mem.NewTable("t1", nil))
+	t2 := newLockableTable(mem.NewTable("t2", nil))
+	t3 := mem.NewTable("t3", nil)
+	catalog := sql.NewCatalog()
+	db := mem.NewDatabase("db")
+	db.AddTable("t1", t1)
+	db.AddTable("t2", t2)
+	db.AddTable("t3", t3)
+	catalog.AddDatabase(db)
+
+	analyzer := analyzer.NewDefault(catalog)
+	engine := sqle.New(catalog, analyzer, new(sqle.Config))
+
+	_, iter, err := engine.Query(newCtx(), "LOCK TABLES t1 READ, t2 WRITE, t3 READ")
+	require.NoError(err)
+
+	_, err = sql.RowIterToRows(iter)
+	require.NoError(err)
+
+	_, iter, err = engine.Query(newCtx(), "UNLOCK TABLES")
+	require.NoError(err)
+
+	_, err = sql.RowIterToRows(iter)
+	require.NoError(err)
+
+	require.Equal(1, t1.readLocks)
+	require.Equal(0, t1.writeLocks)
+	require.Equal(1, t1.unlocks)
+	require.Equal(0, t2.readLocks)
+	require.Equal(1, t2.writeLocks)
+	require.Equal(1, t2.unlocks)
+}
+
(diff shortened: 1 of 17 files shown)


server/handler.go
```diff
@@ -61,6 +61,10 @@ func (h *Handler) ConnectionClosed(c *mysql.Conn) {
 	delete(h.c, c.ConnectionID)
 	h.mu.Unlock()
 
+	if err := h.e.Catalog.UnlockTables(nil, c.ConnectionID); err != nil {
```
mcarmonaa: In this case the tables would keep locked forever?

sql/catalog.go
```diff
@@ -124,3 +133,50 @@ func (d Databases) Table(dbName string, tableName string) (Table, error) {
...
+	defer c.mu.Unlock()
+	if _, ok := c.locks[id]; !ok {
+		c.locks[id] = make(dbLocks)
+	}
+
+	if _, ok := c.locks[id][db]; !ok {
+		c.locks[id][db] = make(tableLocks)
+	}
+
+	c.locks[id][db][table] = struct{}{}
```
kuba--: I'm not sure if we plan (in the future) lock all DB or Session (not only tables, rows), but if yes then maybe flat structure, e.g. `c.locks["id.db.table"]` may work better, because you don't need to create nested structs. Moreover when you unlock you have to propagate this info to the higher levels, instead of delete one key.
But frankly it's up to you.

ajnavarro:
@erizocosmico can you rebase please?

engine_test.go
```diff
@@ -1286,3 +1322,30 @@ func newCtx() *sql.Context {
...
+
+func newLockableTable(t sql.Table) *lockableTable {
+	return &lockableTable{Table: t}
+}
+
+var _ sql.Lockable = (*lockableTable)(nil)
+
+func (l *lockableTable) Lock(ctx *sql.Context, write bool) error {
+	if write {
+		l.writeLocks++
```
kuba--: Why do we count locks?
Isn't it 1/0 (SET/UNSET) for the same session?

sql/parse/parse_test.go
```diff
@@ -788,6 +789,34 @@ var fixtures = map[string]sql.Node{
 		},
 		plan.NewUnresolvedTable("bar", "foo"),
 	),
+	`UNLOCK TABLES`: plan.NewUnlockTables(),
```
ajnavarro: could you test table names with numbers?

engine_test.go
kuba--: ...or is it  just counter for number of read/writes for unlock?

server/handler.go
erizocosmico: If it fails, you mean?

mcarmonaa: yes

engine_test.go
erizocosmico: is just a counter for the test

erizocosmico:
Rebased @ajnavarro

sql/parse/parse_test.go
erizocosmico: Done

server/handler.go
erizocosmico: Then you can retry `UNLOCK TABLES`. If it fails it's because the tables `Unlock` failed, so nothing we can do there, as the implementation of that method depends on the tables.

sql/catalog.go
erizocosmico: The thing is, we use the elements of each separate step. c.locks[id] to get all the locks for a session and then c.locks[id][db] to get the tables. We can't do that with a single key

erizocosmico:
@ajnavarro added to readonly

Issue #444, opened 2018-10-11
Support negative integers in expressions


theodesp:
Currently, any operation that involves negative numbers fail. For example:
```
mysql> SELECT CEIL(-1.23);
ERROR 1105 (HY000): unknown error: unsupported syntax: &sqlparser.UnaryExpr{Operator:"-", Expr:(*sqlparser.SQLVal)(0xc4203cfc60)}
mysql> SELECT FLOOR(-1.23);
ERROR 1105 (HY000): unknown error: unsupported syntax: &sqlparser.UnaryExpr{Operator:"-", Expr:(*sqlparser.SQLVal)(0xc4203cfde0)}
```

PR #456, opened 2018-10-16
sql/(parse,expression): implement unary minus

erizocosmico:
Closes https://github.com/src-d/go-mysql-server/issues/444

### DIFF
--- a/sql/expression/arithmetic.go
+++ b/sql/expression/arithmetic.go
@@ -2,6 +2,7 @@ package expression
 
 import (
 	"fmt"
+	"reflect"
 
 	errors "gopkg.in/src-d/go-errors.v1"
 	"gopkg.in/src-d/go-vitess.v1/vt/sqlparser"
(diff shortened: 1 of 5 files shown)


sql/expression/arithmetic.go
```diff
@@ -407,3 +408,81 @@ func mod(lval, rval interface{}) (interface{}, error) {
...
+	}
+
+	if !sql.IsNumber(e.Child.Type()) {
+		child, err = sql.Float64.Convert(child)
+		if err != nil {
+			child = 0.0
+		}
+	}
+
+	switch n := child.(type) {
```
kuba--: don't we need a case for regular `int`?

kuba--: I think for:
```
var child interface{} = 10
```
it will go to the default.

ajnavarro:
@erizocosmico can you rebase please?

erizocosmico:
Done

ajnavarro:
@erizocosmico could you add a test for:

```
case *sqlparser.UnaryExpr:
	return unaryExprToExpression(v)
```

erizocosmico:
Done

PR #470, opened 2018-10-18
sql: add timing info to span traces

### LINKED ISSUE #466: Traces should be more granular
There should be way more granularity in the traces. Right now, the only meaningful traces are the ones in the analyzer, because they're split per rule.

The execution traces take as much time as the last node, so they don't provide any meaningful insights when looking into which node is the one that's making a query slower.

Perhaps we should measure how much the `Next` in the iterators take as well. I'm worried that this might produce massive traces, though, for big datasets, but it's the only meaningful way to assess what node is making a query slow.

This is blocking https://github.com/src-d/gitbase/issues/560#issuecomment-430391060

WDYT @src-d/data-retrieval?

erizocosmico:
Closes #466

### DIFF
--- a/sql/session.go
+++ b/sql/session.go
@@ -83,6 +83,7 @@ func (s *BaseSession) GetAll() map[string]TypedValue {
 // ID implements the Session interface.
 func (s *BaseSession) ID() uint32 { return s.id }
 
+// TypedValue is a value along with its type.
 type TypedValue struct {
 	Typ   Type
 	Value interface{}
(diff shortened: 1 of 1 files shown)


sql/session.go
```diff
@@ -227,7 +246,13 @@ func (i *spanIter) finish() {
 		LogRecords: []opentracing.LogRecord{
 			{
 				Timestamp: time.Now(),
-				Fields:    []log.Field{log.Int("rows", i.count)},
+				Fields: []log.Field{
+					log.Int("rows", i.count),
+					log.String("total_time", i.total.String()),
+					log.String("max_time", i.max.String()),
+					log.String("min_time", i.min.String()),
+					log.String("avg_time", (i.total / time.Duration(i.count)).String()),
```
kuba--: is it possible here to have _div by 0_?

erizocosmico: yep, you're right

erizocosmico:
Just fixed the div by 0 thing

PR #473, opened 2018-10-18
Fix Apache License

crweiner:
The root of the repo also appears to be missing the Apache v2.0 License. There was a link to the LICENSE file in the README, but no LICENSE file was actually at the repo root.

This PR adds it to the root repo. :+1:


ajnavarro:
Thanks!

PR #482, opened 2018-10-19
Feature/logarithms

theodesp:
PR adds relevant Logarithm functions:
Some examples:

```
mysql> select log();
ERROR 1105 (HY000): unknown error: expecting 1 or 2 arguments for calling this function, 0 received
mysql> select log(1, 2, 3);
ERROR 1105 (HY000): unknown error: expecting 1 or 2 arguments for calling this function, 3 received
mysql> select log(1, 2);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: 1
mysql> select log(-1);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -1
mysql> select log(-1, 100);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -1
mysql> select log(2, 10);
+-------------------+
| log(2, 10)        |
+-------------------+
| 3.321928094887362 |
+-------------------+
1 row in set (0.00 sec)

mysql> select ln(0);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: 0
mysql> select ln(-2);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -2
mysql> select ln(20);
+-------------------+
| ln(20)            |
+-------------------+
| 2.995732273553991 |
+-------------------+
1 row in set (0.00 sec)

mysql> select log2(0);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: 0
mysql> select log2(-2);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -2
mysql> select log2(2);
+---------+
| log2(2) |
+---------+
|       1 |
+---------+
1 row in set (0.00 sec)


mysql> select log2('5');
+-------------------+
| log2("5")         |
+-------------------+
| 2.321928094887362 |
+-------------------+
1 row in set (0.00 sec)

mysql> select log10(-10);
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -10
mysql> select log10('-20');
ERROR 1105 (HY000): unknown error: invalid argument value for logarithm: -20
mysql> select log10(100);
+------------+
| log10(100) |
+------------+
|          2 |
+------------+
1 row in set (0.00 sec)

mysql>
```

### DIFF
--- a/sql/expression/function/logarithm.go
+++ b/sql/expression/function/logarithm.go
@@ -13,41 +13,59 @@ import (
 // logarithm function
 var ErrInvalidArgumentForLogarithm = errors.NewKind("invalid argument value for logarithm: %v")
 
-// Ln is a function that returns the exponential logarithm of a value.
-type Ln struct {
+// LogBaseMaker returns LogBase creator functions with a specific base.
+func LogBaseMaker(base float64) func(e sql.Expression) sql.Expression {
+	b := base
+	return func(e sql.Expression) sql.Expression {
+		return NewLogBase(b, e)
+	}
+}
+
+// LogBase is a function that returns the logarithm of a value with a specific base.
+type LogBase struct {
 	expression.UnaryExpression
+	base float64
 }
 
-// NewLn creates a new Ln expression.
-func NewLn(e sql.Expression) sql.Expression {
-	return &Ln{expression.UnaryExpression{Child: e}}
+// NewLogBase creates a new LogBase expression.
+func NewLogBase(base float64, e sql.Expression) sql.Expression {
+	return &LogBase{UnaryExpression: expression.UnaryExpression{Child: e}, base: base}
 }
 
-func (l *Ln) String() string {
-	return fmt.Sprintf("ln(%s)", l.Child)
+func (l *LogBase) String() string {
+	switch l.base {
+	case float64(math.E):
+		return fmt.Sprintf("ln(%s)", l.Child)
+	case float64(10):
+		return fmt.Sprintf("log10(%s)", l.Child)
+	case float64(2):
+		return fmt.Sprintf("log2(%s)", l.Child)
+	default:
+		return fmt.Sprintf("log(%v, %s)", l.base, l.Child)
+	}
 }
 
 // TransformUp implements the Expression interface.
-func (l *Ln) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
+func (l *LogBase) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
 	child, err := l.Child.TransformUp(f)
 	if err != nil {
 		return nil, err
 	}
-	return f(NewLn(child))
+	return f(NewLogBase(l.base, child))
 }
 
 // Type returns the resultant type of the function.
-func (l *Ln) Type() sql.Type {
+func (l *LogBase) Type() sql.Type {
 	return sql.Float64
 }
 
 // IsNullable implements the sql.Expression interface.
-func (l *Ln) IsNullable() bool {
+func (l *LogBase) IsNullable() bool {
 	return l.Child.IsNullable()
 }
 
 // Eval implements the Expression interface.
-func (l *Ln) Eval(
+func (l *LogBase) Eval(
 	ctx *sql.Context,
 	row sql.Row,
 ) (interface{}, error) {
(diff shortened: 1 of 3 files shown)


sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,313 @@
...
+func (l *Log) Resolved() bool {
+	return l.Left.Resolved() && (l.Right == nil || l.Right.Resolved())
+}
+
+// Eval implements the Expression interface.
+func (l *Log) Eval(
+	ctx *sql.Context,
+	row sql.Row,
+) (interface{}, error) {
+	left, err := l.Left.Eval(ctx, row)
```
kuba--: This common block which repeats multiple times (eval + convert) we can extract to some private function.

kuba--: This common block which repeats multiple times (eval + convert) we can extract to some private function.

kuba--: This common block which repeats multiple times (eval + convert) we can extract to some private function.

sql/expression/function/registry.go
```diff
@@ -43,4 +43,8 @@ var Defaults = sql.Functions{
 	"coalesce":      sql.FunctionN(NewCoalesce),
 	"json_extract":  sql.FunctionN(NewJSONExtract),
 	"connection_id": sql.Function0(NewConnectionID),
+	"ln": 			 sql.Function1(NewLn),
```
kuba--: Do we need all these expressions (Ln, Log2, Log10...)? Maybe we can register all of them (sql functions) as a one `NewLog`, transform them and delegate to one `computeLog` function - wdyt?

sql/expression/function/logarithm.go
theodesp: Sure will push a fix today

theodesp: Sure will push a fix today

sql/expression/function/registry.go
theodesp: Sure we can pass the base as a parameter in NewLog and use it a  guide for the rest of the common methods. Will push a fix today

theodesp:
I've consolidated Log2, Ln and Log10. @kuba-- Please take a look

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Type returns the resultant type of the function.
+func (l *LogBase) Type() sql.Type {
+	return sql.Float64
+}
+
+// IsNullable implements the sql.Expression interface.
+func (l *LogBase) IsNullable() bool {
+	return l.Child.IsNullable()
```
kuba--: IMO Logarithm can be NULL in ranges where function y=log b(x) doesn't exist:
- b or x is NULL
- base is <= 0 or base == 1
- x <= 0

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+	if argLen == 0 || argLen > 2 {
+		return nil, sql.ErrInvalidArgumentNumber.New("1 or 2", argLen)
+	}
+
+	var right sql.Expression
+	if len(args) == 2 {
+		right = args[1]
+	}
+
+	return &Log{expression.BinaryExpression{Left: args[0], Right: right}}, nil
```
kuba--: isn't better to always set `b` and `x`  instead of checking somewhere else which one is nil?
- if len == 1  => left = E, right = args[0]
- if len == 2 => left = args[0], right = args[1]

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
+package function
+
+import (
```
mcarmonaa: About the style in the import block, could you separate the standart libraries from the third libraries by a blank line as below?
```go
import (
        "std1"
        "std2"

        "third1"
        "third2"
)
```

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return f(expr)
+}
+
+// Children implements the Expression interface.
+func (l *Log) Children() []sql.Expression {
+	if l.Right == nil {
```
kuba--: IMO internally we should explicitly say - left is a base, right is argument.
If base is not specified in constructor then we set it up to `E`, but internally we shouldn't have hidden logic, which argument stands for what.

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+	return []sql.Expression{l.Left, l.Right}
+}
+
+// Type returns the resultant type of the function.
+func (l *Log) Type() sql.Type {
+	return sql.Float64
+}
+
+// IsNullable implements the Expression interface.
+func (l *Log) IsNullable() bool { return l.Left.IsNullable() }
```
kuba--: see comments above

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Resolved implements the Expression interface.
+func (l *Log) Resolved() bool {
+	return l.Left.Resolved() && (l.Right == nil || l.Right.Resolved())
```
kuba--: it's only resolvable for arguments from domain

sql/expression/function/logarithm.go
```diff
@@ -0,0 +1,223 @@
...
+	"reflect"
+)
+
+// ErrInvalidArgumentForLogarithm is returned when an invalid argument value is passed to a
+// logarithm function
+var ErrInvalidArgumentForLogarithm = errors.NewKind("invalid argument value for logarithm: %v")
+
+// LogBaseMaker returns LogBase creator functions with a specific base.
+func LogBaseMaker(base float64) func(e sql.Expression) sql.Expression {
+	b := base
```
mcarmonaa: can't you pass directly the argument `base` to the function `NewLogBase` in the closure?

sql/expression/function/logarithm_test.go
```diff
@@ -0,0 +1,198 @@
+package function
+
+import (
```
mcarmonaa: same about import style

sql/expression/function/registry.go
```diff
@@ -3,6 +3,7 @@ package function
 import (
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression/function/aggregation"
+	"math"
```
mcarmonaa: same about import style

sql/expression/function/logarithm.go
theodesp: OK will refactor but I took this logic from 
https://github.com/src-d/go-mysql-server/blob/master/sql/expression/function/ceil_round_floor.go#L155-L166

sql/expression/function/logarithm.go
theodesp: OK I will add the check here also

sql/expression/function/logarithm.go
theodesp: If we setup base to Math.E wouldn't that make difficult to detect if the request was using 1 argument instead of 2. For example when l.Right == nil we know that the request was `log(%s)` but now every time it would be `log(%s, %s)` which looks wierd.

```
mysql> select log(2);
+---------------------------+
| log(2.718281828459045, 2) |
+---------------------------+
|        0.6931471805599453 |
+---------------------------+
```

sql/expression/function/logarithm.go
kuba--: I think it may work.

sql/expression/function/logarithm.go
kuba--: I believe, internally it's better to have everything explicit.

sql/expression/function/logarithm.go
kuba--: But the header comes from the `String` - right?
If yes, maybe we can add extra nice formatting

theodesp: Ok we can just always show the base even if its not specified.

theodesp:
I've simplified the logic even more @kuba-- Please take a look

PR #486, opened 2018-10-21
sql: function, added soundex

BakeRolls:
This PR adds a soundex function as described in the [MySQL Reference](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_soundex) and implemented in [mf_soundex.c](https://github.com/mysql/mysql-server/blob/5.7/mysys/mf_soundex.c) in mysql/mysql-server.


sql/expression/function/soundex.go
```diff
@@ -0,0 +1,102 @@
...
+		if code == '0' || code == last {
+			continue
+		}
+		b.WriteRune(code)
+		last = code
+	}
+	if b.Len() == 0 {
+		return "", nil
+	}
+	for i := len([]rune(b.String())); i < 4; i++ {
```
kuba--: why 4 (because type rune = int32)?

kuba--: why 4 (because type rune = int32)?

BakeRolls: If a given string can be represented with soundex (if it has at lease one character listed in the `code` function), the resulting string can have an arbitrary length, but at least four characters:

```
mysql> select soundex('');
+-------------+
| soundex('') |
+-------------+
|             |
+-------------+
1 row in set (0,00 sec)

mysql> select soundex('Q');
+--------------+
| soundex('Q') |
+--------------+
| Q000         |
+--------------+
1 row in set (0,01 sec)

mysql> select soundex('Quadratically');
+--------------------------+
| soundex('Quadratically') |
+--------------------------+
| Q36324                   |
+--------------------------+
1 row in set (0,00 sec)
```

BakeRolls: (I am not sure if my previous comment got saved)

Even though the reference doesn't say so, MySQL returns at least four characters, if one of them is a letter:

```
mysql> select soundex('');
+-------------+
| soundex('') |
+-------------+
|             |
+-------------+
1 row in set (0,00 sec)

mysql> select soundex('Q');
+--------------+
| soundex('Q') |
+--------------+
| Q000         |
+--------------+
1 row in set (0,00 sec)

mysql> select soundex('Quadratically');
+--------------------------+
| soundex('Quadratically') |
+--------------------------+
| Q36324                   |
+--------------------------+
1 row in set (0,00 sec)
```

BakeRolls: I am not sure how close this should follow MySQL, since it is not a replacement. Always returning a string of 0 or 4 characters would be a possibility, too.

BakeRolls:
@kuba-- Should I change something?

kuba--:
@BakeRolls - could you rebase. Btw. I accepted this PR but to merge it we have to wait for a maintainer.

BakeRolls:
Thank you!

ajnavarro:
@BakeRolls thanks for the contribution!

PR #498, opened 2018-10-24
analyzer: fix pushdown of filters with repeated table names

erizocosmico:
Fixes https://github.com/src-d/gitbase/issues/579

### DIFF
--- a/sql/analyzer/filters.go
+++ b/sql/analyzer/filters.go
@@ -18,18 +18,22 @@ func (f filters) merge(f2 filters) {
 func exprToTableFilters(expr sql.Expression) filters {
 	filtersByTable := make(filters)
 	for _, expr := range splitExpression(expr) {
-		var tables []string
+		var seenTables = make(map[string]struct{})
+		var lastTable string
 		_, _ = expr.TransformUp(func(e sql.Expression) (sql.Expression, error) {
 			f, ok := e.(*expression.GetField)
 			if ok {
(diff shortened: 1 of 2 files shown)


sql/analyzer/filters.go
```diff
@@ -18,11 +18,15 @@ func (f filters) merge(f2 filters) {
 func exprToTableFilters(expr sql.Expression) filters {
 	filtersByTable := make(filters)
 	for _, expr := range splitExpression(expr) {
+		var seenTables = make(map[string]struct{})
 		var tables []string
```
kuba--: do we need `var tables []string` now? If we just check:
```
if len(tables) == 1 {
 			filtersByTable[tables[0]] = append(filtersByTable[tables[0]], expr)		 			filtersByTable[tables[0]] = append(filtersByTable[tables[0]], expr)
 		}		 		}
```
?

erizocosmico: How do you get the key from the map, then? You would need to range the map and that feels like this with extra steps

kuba--: that's true - we'll need to `range` and break, but on the other hand if we really need just 0th element then we may just memoize just one string instead of all slice and only set it `if len(seen) == 0`

erizocosmico: Fixed

erizocosmico:
Can we merge this before I update gitbase, so it also gets this update? @ajnavarro

PR #499, opened 2018-10-24
Add support for CONCAT_WS

BakeRolls:
Add support for [`CONCAT_WS`](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_concat-ws). Other than `CONCAT`, it returns `null` only if the separator is `null`. `null` fields are skipped.


BakeRolls:
Sorry, added wrong files to the branch.

PR #500, opened 2018-10-24
Add support for CONCAT_WS

BakeRolls:
Add support for [`CONCAT_WS`](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_concat-ws). Other than `CONCAT`, it returns `null` only if the separator is `null`. `null` fields are skipped.


README.md
```diff
@@ -56,11 +56,12 @@ We support and actively test against certain third-party clients to ensure compa
 ## Custom functions
 
 - `IS_BINARY(blob)`: Returns whether a BLOB is a binary file or not.
-- `SUBSTRING(str,pos)`, `SUBSTRING(str,pos,len)`: Return a substring from the provided string.
+- `SUBSTRING(str, pos)`, `SUBSTRING(str, pos, len)`: Return a substring from the provided string.
 - Date and Timestamp functions: `YEAR(date)`, `MONTH(date)`, `DAY(date)`, `HOUR(date)`, `MINUTE(date)`, `SECOND(date)`, `DAYOFYEAR(date)`.
 - `ARRAY_LENGTH(json)`: If the json representation is an array, this function returns its size.
 - `SPLIT(str,sep)`: Receives a string and a separator and returns the parts of the string split by the separator as a JSON array of strings.
 - `CONCAT(...)`: Concatenate any group of fields into a single string.
+- `CONCAT_WS(sep, ...)`: Concatenate any group of fields into a single string separated by the first field.
```
erizocosmico: Probably worth mentioning that null arguments are skipped and only returns null if sep is null.

BakeRolls: Thank you!
"Returns null if the separator is null. Following null fields are skipped." - Would this be ok?

erizocosmico: Sounds good to me!

kuba--: Lets copy paste from mysql docs:
>... stands for Concatenate With Separator and is a special form of CONCAT(). The first argument is the separator for the rest of the arguments. The separator is added between the strings to be concatenated. The separator can be a string, as can the rest of the arguments. If the separator is NULL, the result is NULL.

kuba--:
Please sign a PR.

README.md
BakeRolls: Sorry, I only just saw the comment. The descriptions now come from MySQLs docs.

PR #508, opened 2018-10-26
added missing spaces in comments

Julineo:
Edited comments, added missing spaces.


Julineo:
Hi, how can I label this pull request with hacktoberfest label?

campoy:
Hi Alexey,

First of all thanks for the contribution! We love to have first time contributors ❤️ 

Regarding your fix, the name of the interface is actually [opentracing.Tracer](https://godoc.org/github.com/opentracing/opentracing-go#Tracer), so the space is actually not missing.

Also, i think you mistakenly sent your vim `swap` file 😄

PR #513, opened 2018-10-30
added some change in database access

ShobhitSrivastava18th:


ajnavarro:
If you want to add some library compatibilities examples, you can do it here:

https://github.com/src-d/go-mysql-server/blob/master/SUPPORTED_CLIENTS.md

To be able to add as a supported client here, you must add an integration test. You have some examples here: https://github.com/src-d/go-mysql-server/tree/master/_integration

PR #516, opened 2018-10-31
Update README.md

vyaspranjal33:
removed LICENSE badge 

please see this https://github.com/src-d/enry/pull/176#issuecomment-434157019


erizocosmico:
That's a different project and thus may have different README. We do have badges in this one.

PR #526, opened 2018-11-02
sql/analyzer: add rule to avoid unnecessary casts

### LINKED ISSUE #502: Rule to remove unnecessary CAST nodes
For example, if something is of type BLOB and is wrapped in `CAST(foo AS BLOB)`, convert it to just `foo`

erizocosmico:
Fixes #502

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -32,6 +32,7 @@ var OnceBeforeDefault = []Rule{
 // OnceAfterDefault contains the rules to be applied just once after the
 // DefaultRules.
 var OnceAfterDefault = []Rule{
+	{"remove_unnecessary_converts", removeUnnecessaryConverts},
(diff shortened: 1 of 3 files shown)


sql/analyzer/rules.go
```diff
@@ -20,6 +20,7 @@ var DefaultRules = []Rule{
 	{"move_join_conds_to_filter", moveJoinConditionsToFilter},
 	{"eval_filter", evalFilter},
 	{"optimize_distinct", optimizeDistinct},
+	{"remove_unnecessary_converts", removeUnnecessaryConverts},
```
mcarmonaa: What do you think about moving this rule to `OnceAfterDefault`? Is it necessary run it several times?

ajnavarro: @mcarmonaa It should be executed several times because the plan will change and could contain new unnecessary casts again.

PR #534, opened 2018-11-06
Feature/pow sqrt

theodesp:


sql/expression/function/sqrt_power.go
```diff
@@ -0,0 +1,162 @@
...
+	if child == nil {
+		return nil, nil
+	}
+
+	child, err = sql.Float64.Convert(child)
+	if err != nil {
+		return nil, nil
+	}
+
+	return computeSqrt(child.(float64))
```
mcarmonaa: maybe you could use here directly `math.Sqrt` an avoid the extra function call.

sql/expression/function/sqrt_power.go
```diff
@@ -0,0 +1,162 @@
...
+	if right == nil {
+		return nil, nil
+	}
+
+	right, err = sql.Float64.Convert(right)
+	if err != nil {
+		return nil, err
+	}
+
+	return computePower(left.(float64), right.(float64))
```
mcarmonaa: Same as with sqrt, you could use directly `math.Pow` here.

sql/expression/function/sqrt_power.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return computeSqrt(child.(float64))
+}
+
+func computeSqrt(x float64) (float64, error) {
+	return math.Sqrt(x), nil
+}
+
+type powerFuncName int
```
mcarmonaa: Since `power` is a synonym for `pow` (https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_pow) I think is not necessary have different constructors depending on the name.

We could have just `NewPower` without `powerFuncName`s and use it for both functions where they are registered.

theodesp:
Pr updated

sql/expression/function/sqrt_power.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return math.Sqrt(child.(float64)), nil
+}
+
+// Power is a function that returns value of X raised to the power of Y.
+type Power struct {
+	expression.BinaryExpression
+}
+
+// NewPad creates a new Power expression.
```
mcarmonaa: This comment doesn't match the function name.

PR #543, opened 2018-11-12
Add Reverse, Repeat, Replace

theodesp:
Signed-off-by: Theo Despoudis <thdespou@hotmail.com>

### DIFF
--- a/sql/expression/function/reverse_repeat_replace.go
+++ b/sql/expression/function/reverse_repeat_replace.go
@@ -6,7 +6,6 @@ import (
 
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
-
 	"gopkg.in/src-d/go-errors.v1"
 )
 
--- a/sql/expression/function/reverse_repeat_replace.go
+++ b/sql/expression/function/reverse_repeat_replace.go
@@ -14,7 +14,7 @@ type Reverse struct {
 	expression.UnaryExpression
 }
 
-// NewLower creates a new Lower expression.
+// NewReverse creates a new Reverse expression.
 func NewReverse(e sql.Expression) sql.Expression {
 	return &Reverse{expression.UnaryExpression{Child: e}}
 }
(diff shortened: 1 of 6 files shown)


sql/expression/function/reverse_repeat_replace.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return reverseString(v.(string)), nil
+}
+
+func reverseString(s string) string {
+	r := []rune(s)
+	for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 {
```
kuba--: I think it's better to have a `j < j` condition instead of calculating every time length of string and divide by 2.

sql/expression/function/reverse_repeat_replace.go
```diff
@@ -0,0 +1,253 @@
...
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-errors.v1"
+)
+
+// Reverse is a function that returns the reverse of the text provided.
+type Reverse struct {
+	expression.UnaryExpression
+}
+
+// NewLower creates a new Lower expression.
```
jfontan: Comment should match the function.

sql/expression/function/reverse_repeat_replace.go
```diff
@@ -0,0 +1,253 @@
...
+	return &Reverse{expression.UnaryExpression{Child: e}}
+}
+
+// Eval implements the Expression interface.
+func (r *Reverse) Eval(
+	ctx *sql.Context,
+	row sql.Row,
+) (interface{}, error) {
+	v, err := r.Child.Eval(ctx, row)
+	if err != nil {
```
jfontan: This is not wrong but you can deal both cases of `nil` and error with one statement:

```go
if v == nil || err != nil {
      return nil, err
}
```

sql/expression/function/reverse_repeat_replace.go
```diff
@@ -44,7 +44,9 @@ func (r *Reverse) Eval(
 
 func reverseString(s string) string {
 	r := []rune(s)
-	for i, j := 0, len(r)-1; i < len(r)/2; i, j = i+1, j-1 {
+	l := len(r)
```
mcarmonaa: I guess what @kuba-- commented was about using `i < j` directly as the condition for the `for` loop, you don't need to calculate `l` and `l2`

```go
for i, j := 0, len(r) - 1; i < j; i, j = i+1, j-1 {
        r[i], r[j] = r[j], r[i]
}
```

ajnavarro:
@theodesp please, could you rebase to fix CI? thanks.

PR #560, opened 2018-11-28
Fix soundex type

ajnavarro:
Signed-off-by: Antonio Jesus Navarro Perez <antnavper@gmail.com>


ajnavarro:
@mcarmonaa should be fixed now. I checked the soundex correct output when no match, and it is "0000" instead of ""

PR #569, opened 2018-12-05
*: reset proces progress on partition iteration

erizocosmico:
Fixes https://github.com/src-d/gitbase/issues/646 (kinda)

The problem is that it's in an InnerJoin, so one of the branches is **fully** iterated N times, where N is the number of rows in the other branch.

So, the maximum allowed progress would be NxM, being M the number of repositories.

There's no way to know N ahead of time so we can't estimate correctly. The only viable "solution" is to reset the progress in each iteration :/

Idk what you guys think about this, but the only reasonable thing I can think of.


ajnavarro:
I think we should think more the solution for the main bug. As it is not a really high priority bug, I would say to keep as it is until we find a better solution for that. WDYT @src-d/data-retrieval ?

erizocosmico:
Sounds good

PR #576, opened 2018-12-21
sql/(analyzer,parse,expression): implement case expression

### LINKED ISSUE #575: CASE statement
We use `CASE WHEN x THEN y` statements inside gitbase-spark-connector, but it's not actually supported inside gitbase, so some queries that use it may fail.

erizocosmico:
Closes #575

### DIFF
--- a/sql/analyzer/validation_rules.go
+++ b/sql/analyzer/validation_rules.go
@@ -10,12 +10,13 @@ import (
 )
 
 const (
-	validateResolvedRule      = "validate_resolved"
-	validateOrderByRule       = "validate_order_by"
-	validateGroupByRule       = "validate_group_by"
-	validateSchemaSourceRule  = "validate_schema_source"
-	validateProjectTuplesRule = "validate_project_tuples"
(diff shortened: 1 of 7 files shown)


sql/analyzer/validation_rules.go
```diff
@@ -46,6 +53,7 @@ var DefaultValidationRules = []Rule{
 	{validateSchemaSourceRule, validateSchemaSource},
 	{validateProjectTuplesRule, validateProjectTuples},
 	{validateIndexCreationRule, validateIndexCreation},
+	{validateCaseResultTypesRule, validateCaseResultTypes},
```
kuba--: Just open question - doesn't make sense as a optimization to squash some rules?
For instance instead of 3 times looping, we may have one loop with some cases - loud thinking.

ajnavarro: this query phase is really fast compared with the rest. The idea of having separated rules is the synergy that we can obtain executing them again and again, interacting with other rules.

ajnavarro:
@erizocosmico could you rebase please?

PR #578, opened 2018-12-21
sql/plan: speed up InnerJoin iterator

erizocosmico:
```
BenchmarkInnerJoin/inner_join-4         	   50000	     34188 ns/op	   11415 B/op	     169 allocs/op
BenchmarkInnerJoin/cross_join_with_filter-4         	   30000	     41608 ns/op	   12119 B/op	     181 allocs/op
PASS
ok  	gopkg.in/src-d/go-mysql-server.v0/sql/plan	3.795s
```

### DIFF
--- a/sql/plan/innerjoin.go
+++ b/sql/plan/innerjoin.go
@@ -1,6 +1,7 @@
 package plan
 
 import (
+	"io"
 	"reflect"
 
 	opentracing "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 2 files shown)


sql/plan/innerjoin.go
```diff
@@ -127,3 +125,74 @@ func (j *InnerJoin) TransformExpressions(f sql.TransformExprFunc) (sql.Node, err
...
+			i.r = nil
+			i.leftRow = nil
+			continue
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		var row sql.Row
```
kuba--: actually it looks like we can just write:
```
i.leftRow = append(i.leftRow, rightRow...)
```
am I wrong?

kuba--: and eval at the end

erizocosmico:
Sure @ajnavarro

sql/plan/innerjoin.go
erizocosmico: We cannot append rightRow to leftRow.

When we iterate we iterate the left side only once and for each left row, we get many right rows.

So the result row is Left row + Right row.

If we do 
```i.leftRow = append(i.leftRow, rightRow...)```

Left row would change each iteration, which is not what we want.

erizocosmico:
Done

PR #583, opened 2019-01-10
sql: use checksums to determine if indexes are outdated

erizocosmico:
First part of https://github.com/src-d/gitbase/issues/661

After a few attempts, I've settled with this way of handling this, since this is backwards compatible, does not introduce any breaking changes and uses existing mechanisms.

Checksum is sent to the driver using the index configuration that we were not using for anything yet. Instead of changing the API of the drivers to accept a checksum, since this is more of a temporal solution until updates are supported, I preferred to use this way.

Indexes hold the checksum value inside so they can return it. An `sql.Index` may or may not implement `sql.Checksumable`, which makes it really up to the driver to use checksums or not.

Tables may also implement `sql.Checksumable`. If a table is not checksumable, the driver will do nothing related to checksums. If it is, when the driver is loaded it will compare both checksums and mark the index as outdated if they don't match.

A new `IndexOutdated` status has been added for indexes. The index is loaded, so it can be dropped, but it can not be used in any query or overwritten before deleting it. Not loading it if it was outdated would have made it possible for multiple identical indexes to exist at the same time for different versions, which did not make much sense.
When the index is marked as outdated a warning is printed.

### DIFF
--- a/sql/index.go
+++ b/sql/index.go
@@ -5,11 +5,22 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/sirupsen/logrus"
 	"gopkg.in/src-d/go-errors.v1"
 )
 
 // IndexBatchSize is the number of rows to save at a time when creating indexes.
 const IndexBatchSize = uint64(10000)
+
+// ChecksumKey is the key in an index config to store the checksum.
+const ChecksumKey = "checksum"
+
+// Checksumable provides the checksum of some data.
+type Checksumable interface {
+	// Checksum returns a checksum and an error if there was any problem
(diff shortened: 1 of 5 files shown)


erizocosmico:
@kuba-- since this is temporary until we support updates I went the fast path: if the repo changes, the index is outdated. Checking if something has changed for an expression would essentially be recreating the index and comparing, which is not feasible.

kuba--:
Sometimes I have a feeling, that maybe we should trigger two go routines to find values (one with index and the second one will ignore the index). Generally the first one wins, but in case of error, we can wait for the second one).

PR #586, opened 2019-01-15
sql/index/pilosa: return all index ids in lookups

### LINKED ISSUE #585: EXPLAIN only shows the first index
I created 3 indexes:
```
CREATE INDEX a ON commit_files USING pilosa (commit_hash);
CREATE INDEX b ON commit_files USING pilosa (file_path);
CREATE INDEX c ON commit_files USING pilosa (blob_hash);
```

And then tried to show explain for query:
```
EXPLAIN FORMAT=TREE SELECT *
FROM commit_files
WHERE 
commit_hash = 'd5b04a971262372a1531119d595cd4b04db2949a' AND 
file_path = '.gitignore' AND 
blob_hash = 'a7904a1efc6ef7d25b358c0f7ee33259fb4512e0';
```

`EXPLAIN` shows that only index `'a'` will be used (but all 3 indexes should be used).
I suppose it's because of the function `indexesIntersection` in `assign_indexes.go`:
```go
//...
for table, idx := range left {
		if idx2, ok := right[table]; ok && canMergeIndexes(idx.lookup, idx2.lookup) {
			idx.lookup = idx.lookup.(sql.SetOperations).Intersection(idx2.lookup)
			idx.indexes = append(idx.indexes, idx2.indexes...)
		} 
//...
```
where we merge index `b` into th `a` and then index `c` into `intersection of a & b`. At the end we have one _lookup_ `a` merged with 2 operations (b, c). Moreover the name is still `a` so explain shows only `a`.
The same happens for other operations (not only for AND). We can change order of `WHERE` conditions to see that always the first one is showed by `explain`

erizocosmico:
Fixes #585

### DIFF
--- a/sql/index/pilosa/lookup.go
+++ b/sql/index/pilosa/lookup.go
@@ -64,7 +64,7 @@ type (
 		keys        []interface{}
 		expressions []string
 		operations  []*lookupOperation
-		indexes     []string
+		indexes     map[string]struct{}
 	}
 
 	lookupOperation struct {
@@ -167,7 +167,7 @@ func (l *indexLookup) Values(p sql.Partition) (sql.IndexValueIter, error) {
 }
 
 func (l *indexLookup) Indexes() []string {
-	return uniqueIndexes(append(l.indexes[:], l.id))
+	return sortedIndexes(l.indexes)
 }
 
 // IsMergeable implements sql.Mergeable interface.
@@ -183,7 +183,9 @@ func (l *indexLookup) IsMergeable(lookup sql.IndexLookup) bool {
 func (l *indexLookup) Intersection(lookups ...sql.IndexLookup) sql.IndexLookup {
 	lookup := *l
 	for _, li := range lookups {
-		lookup.indexes = append(lookup.indexes, li.Indexes()...)
+		for _, idx := range li.Indexes() {
+			lookup.indexes[idx] = struct{}{}
+		}
 		lookup.operations = append(lookup.operations, &lookupOperation{li, intersect})
 	}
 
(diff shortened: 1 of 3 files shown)


sql/index/pilosa/lookup.go
```diff
@@ -63,6 +64,7 @@ type (
 		keys        []interface{}
 		expressions []string
 		operations  []*lookupOperation
+		indexes     []string
```
kuba--: Maybe it should be `map[string]int`, where key is an index name and value will be order id.
So we'll have auto-uniques (because of map), and then we can sort map entries by value, sth. like:
```go
m := make(map[string]int)
m[indexName] = len(m)
//...
indexes := make([]string, len(m))
for k, v := range m {
    index[k] = v
}
return index
```

WDYT?

erizocosmico: Why the `int`? We can just use a `map[string]struct{}` and then sort them.

Also, even if `indexes` is a map, it may not be unique, because `id` can be inside that map.

kuba--: The question is how do we want to show indexes (sorted in the order how filter was parsed, or sorted by name)?
On the other hand, these are "small" numbers so most likely keep the most readable version.

erizocosmico: I will use the `struct{}` only because the int requires some explanation at first sight and the other doesn't. Regarding the order, I think it does not matter much, so we can just sort.

kuba--: Ok, so let's keep it what we have so far. I suppose, there is no many cases where we'll use the same index more than once, so maybe map can be kind of overhead.

erizocosmico:
Done, though I think it unnecessarily complicates the code for very little gain.

kuba--:
Btw, we have to fix old "merge" tests: https://github.com/src-d/go-mysql-server/pull/586/files#diff-136ffc4118af2fd83a7cbf2bc27fe357R215

PR #588, opened 2019-01-17
sql/*: fix show tables to match MySQL spec

### LINKED ISSUE #562: [SQLAlchemy] missing column in SHOW FULL TABLES

SQLAlchemy uses `Table_type` column in `SHOW FULL TABLES` results:
   * [get_table_names](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/mysql/base.py#L2069)  and line  [:2081](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/mysql/base.py#L2081)

> As of MySQL 5.0.2, the FULL modifier is supported such that SHOW FULL TABLES displays a second output column
   http://ftp.nchu.edu.tw/MySQL/doc/refman/5.0/en/show-tables.html

Right now `SHOW FULL TABLES` works, but does not include a `Table_type` column.

erizocosmico:
Fixes #562 

- Implements all what's in the spec for `SHOW TABLES`, which caused an error before even if we had show tables implemented.
- Fixes a bug on the resolve_database analyzer rule.

### DIFF
--- a/sql/analyzer/resolve_database.go
+++ b/sql/analyzer/resolve_database.go
@@ -11,53 +11,60 @@ func resolveDatabase(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error
 
 	a.Log("resolve database, node of type: %T", n)
 
-	switch v := n.(type) {
-	case *plan.ShowIndexes:
-		db, err := a.Catalog.Database(a.Catalog.CurrentDatabase())
-		if err != nil {
-			return nil, err
-		}
+	return n.TransformUp(func(n sql.Node) (sql.Node, error) {
+		switch v := n.(type) {
+		case *plan.ShowIndexes:
+			db, err := a.Catalog.Database(a.Catalog.CurrentDatabase())
+			if err != nil {
+				return nil, err
+			}
 
-		nc := *v
-		nc.Database = db
-		return &nc, nil
-	case *plan.ShowTables:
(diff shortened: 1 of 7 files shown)


sql/analyzer/resolve_database.go
```diff
@@ -11,53 +11,60 @@ func resolveDatabase(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error
...
+				return nil, err
+			}
+
+			nc := *v
+			nc.Database = db
+			return &nc, nil
+		case *plan.ShowCreateDatabase:
+			db, err := a.Catalog.Database(v.Database.Name())
+			if err != nil {
+				return nil, err
```
kuba--: This kind of block is almost in every case. Maybe it's good time to extract it to some function.

erizocosmico: But the type is different, so you can't extract it. We could extract the `db, err := a.Catalog.Database(v.Database.Name())`, but after error handling it would be the same amount of lines, so no point in extracting that either.

The only way I can imagine that we can use to extract this is create an interface that these nodes implement.

```go
type WithDatabaser interface {
  WithDatabase(sql.Database) (sql.Node, error)
}
```

kuba--: ohh, you're right - it sucks. Let's leave it.

erizocosmico: #generics 😜

PR #589, opened 2019-01-17
sql: add transaction_isolation default value

### LINKED ISSUE #564: [SQLAlchemy] SELECT @@transaction_isolation support or workaround

SQLAlchemy uses `SELECT @@transaction_isolation` in:
   * [get_isolation_level](thttps://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/mysql/base.py#L1819)
   
Most probably this is relevant to #563 and could be somehow worked around (returning value that is not `None`?), but a workaround is unknown.

Here is a [python snippet](https://gist.github.com/bzz/94c43cb22fa2bf69c09ccb4e47cfe781#python-code) to reproduce this.

erizocosmico:
Fixes #564 

I added `READ UNCOMMITED` because it's what's most similar to what we do (no transactions at all).

> `READ UNCOMMITTED`
> 
> SELECT statements are performed in a nonlocking fashion, but a possible earlier version of a row > might be used. Thus, using this isolation level, such reads are not consistent. This is also called a dirty read. Otherwise, this isolation level works like READ COMMITTED.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -542,6 +542,7 @@ var queries = []struct {
 			{"collation_database", "utf8_bin"},
 			{"ndbinfo_version", ""},
 			{"sql_select_limit", math.MaxInt32},
+			{"transaction_isolation", "READ UNCOMMITTED"},
 		},
 	},
 	{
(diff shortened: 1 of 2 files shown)


erizocosmico:
Fixed @kuba--

PR #593, opened 2019-01-18
sql: implement dummy ROLLBACK

### LINKED ISSUE #563: [SQLAlchemy] ROLLBACK support or workaround

SQLAlchemy sends `ROLLBACK` quite often:
   * [engine/default.py#initialize](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L285)
   * on recovery from any failure
   
This could possibly be worked around though [connection configuration options](https://github.com/zzzeek/sqlalchemy/blob/c99345ee9994c3ea2a5e6536cc3365f18d017cc1/doc/build/faq/connections.rst#why-does-sqlalchemy-issue-so-many-rollbacks), but a workaround is unknown.

Here is a [python snippet](https://gist.github.com/bzz/94c43cb22fa2bf69c09ccb4e47cfe781#python-code) to reproduce this.

erizocosmico:
Closes #563

### DIFF
--- a/sql/plan/transaction.go
+++ b/sql/plan/transaction.go
@@ -1,0 +1,35 @@
+package plan
+
+import "gopkg.in/src-d/go-mysql-server.v0/sql"
+
+// Rollback undoes the changes perfromed in a transaction.
+type Rollback struct{}
+
+// NewRollback creates a new Rollback node.
+func NewRollback() *Rollback { return new(Rollback) }
+
+// RowIter implements the sql.Node interface.
+func (*Rollback) RowIter(*sql.Context) (sql.RowIter, error) {
+	return sql.RowsToRowIter(), nil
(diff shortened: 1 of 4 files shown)


sql/plan/transaction.go
```diff
@@ -0,0 +1,35 @@
+package plan
+
+import "gopkg.in/src-d/go-mysql-server.v0/sql"
+
+// Rollback undoes the changes perfromed in a transaction.
```
jfontan:
```suggestion
// Rollback undoes the changes performed in a transaction.
```

erizocosmico:
@kuba-- yup, just for it not to fail. Some clients to a lot of `ROLLBACK`s, such as SQLAlchemy

PR #596, opened 2019-01-18
sql: better errors for invalid argument number

### LINKED ISSUE #591: Improve Function invalid argument number error specifying the affected function name
The following query
```
select 
    language(blob_content, file_path) as lang, 
    uast_extract(
        uast_xpath(
            uast_xpath(
                uast_mode('annotated', blob_content, lang)
               '//InfixExpression[@role="Add" and @role="Binary" and @role="Operator"]'
            ),
            "//String[@internalRole='leftOperand']"),
            "Value"
        ) as result 
from refs 
NATURAL JOIN commits 
NATURAL JOIN commit_files 
NATURAL JOIN files 
where lang != '' limit 100;
```
Is failing with `ERROR 1105 (HY000): unknown error: expecting 2 arguments for calling this function, 1 received`

We should specify the function name on the error at least.

erizocosmico:
Fixes #591

### DIFF
--- a/sql/expression/function/ceil_round_floor.go
+++ b/sql/expression/function/ceil_round_floor.go
@@ -155,7 +155,7 @@ type Round struct {
 func NewRound(args ...sql.Expression) (sql.Expression, error) {
 	argLen := len(args)
 	if argLen == 0 || argLen > 2 {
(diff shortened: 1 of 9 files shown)


ajnavarro:
@kuba-- you cannot use the String method before having the instance on the constructor, right?

PR #599, opened 2019-01-21
Add support for HAVING clauses

### LINKED ISSUE #56: TPC-H: unsupported feature: HAVING
Implement HAVING functionality: https://www.w3schools.com/sql/sql_having.asp

andremarianiello:
fixes #56


andremarianiello:
I saw that HAVING clauses were not supported (https://github.com/src-d/go-mysql-server/issues/56) and I would really like them to be supported. The changes I made now allow you to write, for the example:

```sql
MySQL [(none)]> select name, count(email) as num_emails from mytable group by name having num_emails > 1;
+----------+------------+
| name     | num_emails |
+----------+------------+
| John Doe |          2 |
+----------+------------+
1 row in set (0.000 sec)
```

However, this took like 5 seconds to do, so I'm wondering how I should find out if this is working as intended? There are no tests for HAVING, as expected, so I can add some if the reviewer(s) think this is a valid approach to implementing having.

erizocosmico:
Hi @andremarianiello, sorry for the waiting. First of all, thanks for the contribution!

It actually looks good to me. I'd just ask you to add some tests:

- In `sql/parse/parse_test.go` to ensure the parsing is correct.
- In `engine_test.go` as an integration test to ensure HAVING works end to end. This one is specially important because synergies between analyzer rules may do some magic to filter nodes and we might have to create a special node (even if it's just the same as a filter) for this. But let's hope it just works 😛

andremarianiello:
I have added a simple parsing test for HAVING. However, I have run into a bit of an issue for the end-to-end test. The root of the problem is when I try the query:

```sql
SELECT substring(mytable.s, 1, 5) as s 
FROM mytable INNER JOIN othertable ON (substring(mytable.s, 1, 5) = SUBSTRING(othertable.s2, 1, 5))
GROUP BY 1 HAVING s = "secon"
```

the `s` column in the HAVING clause is being qualified with `mytable`. This causes a confusing error message

```
table "mytable" does not have column "s"
```

I was trying to refer to the alias "s", not the column "mytable.s". If I use a non-colliding alias name, everything works fine. I changed the logic in qualifyColumns to treat unresolved column expressions in a HAVING node (i.e. a filter with a groupby child) the same way they are treated in groupby nodes. Please let me know if that logic is off.

erizocosmico:
I’m a little bit concerned about assuming that any Filter node with a GroupBy node is a Having node (it’s always true after parsing, but may change in analysis phase due to some rule and no longer be true). If we reuse Filter node for having there is no way to tell which is which.

I think the best option is just going to be adding a new Having node, even if it internally reuses the filter iterator, just to always be able to differentiate between a having and a filter and then just add Having to the check in 

Other than that, LGTM.

andremarianiello:
I'm more concerned about the logic for treating GroupBy specially when qualifying column names. Can anyone explain why it currently does that? Ideally we could figure out a solution that avoid special-casing both GroupBy and Having (thus not requiring a Having node, at least for for this purpose).

erizocosmico:
@andremarianiello because in projections we don't allow usage of a column defined in that same node, but in GroupBy we do (grouping clause), which should not be happening on a filter node.

Can you print the node at the moment of the error and post the resultant tree here? Maybe it's because of some movement of the filter node.

erizocosmico:
Yeah, so the issue here is that it should never qualify s as `mytable.s`, the order seems to be right. It should only qualify using whatever's available on the child nodes instead of all available stuff across the tree.

andremarianiello:
@erizocosmico Do you have a specific way you want to handle this, or should I keep looking into it?

erizocosmico:
@andremarianiello basically, when getting all possible table-column mappings in `qualifyColumns` it should use for each node only what's projected in the children. If there is more than one child, then use the concatenation of their schemas (such as inner join and so on). I think that should do it.

andremarianiello:
In the interest of not expanding the scope of this issue to be more than it should be, can I add HAVING support and HAVING test cases as part of this PR, and save the fixes for the alias issue for another issue?

ajnavarro:
We prefer not to add new syntax until we are sure that its behavior is reasonably correct and does not introduce new known bugs.

Don't hesitate to ask any question or doubt about new ideas of how to implement this.

Thanks a lot.

erizocosmico:
Closing, this is being addressed in #689

PR #603, opened 2019-01-23
sql/analyzer: remove QueryProcess nodes from subqueries

erizocosmico:
Closes https://github.com/src-d/gitbase/issues/679

Because subqueries are analyzed as if they were a whole independent
query QueryProcess nodes were added to them. That caused that as
soon as a subquery finished, the process were marked as done. As a
result, some queries could error with "context canceled", as the
context of the query had already been canceled.

This fix removes the QueryProcess nodes from any subquery and makes
sure only the root node is wrapped in a QueryProcess.


erizocosmico:
Merging this because it's blocking a task for the PoC.

PR #605, opened 2019-01-28
add experimental optional feature to use in memory joins

erizocosmico:
This exposes the env var `EXPERIMENTAL_IN_MEMORY_JOIN` to enable the in-memory join feature, which causes inner joins to be performed in memory, which is significantly faster.

This way, clients can decide if they should enable in memory joins or not depending on their environment. If they can trade memory usage for speed this is orders of magnitude faster (specially for computationally expensive inner join branches) than the regular inner joins.

These are the benchmarks:

```
goos: darwin
goarch: amd64
BenchmarkInnerJoin/inner_join-4         	   50000	     33170 ns/op	   10618 B/op	     144 allocs/op
BenchmarkInnerJoin/in_memory_inner_join-4         	  100000	     21972 ns/op	    7322 B/op	      84 allocs/op
BenchmarkInnerJoin/cross_join_with_filter-4       	   30000	     41750 ns/op	   12122 B/op	     181 allocs/op
PASS
ok  	gopkg.in/src-d/go-mysql-server.v0/sql/plan	6.112s
```

### DIFF
--- a/sql/plan/innerjoin.go
+++ b/sql/plan/innerjoin.go
@@ -2,11 +2,16 @@ package plan
 
 import (
 	"io"
+	"os"
 	"reflect"
 
 	opentracing "github.com/opentracing/opentracing-go"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
+
+const experimentalInMemoryJoinKey = "EXPERIMENTAL_IN_MEMORY_JOIN"
+
+var useInMemoryJoins = os.Getenv(experimentalInMemoryJoinKey) != ""
 
 // InnerJoin is an inner join between two tables.
 type InnerJoin struct {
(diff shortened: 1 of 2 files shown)


erizocosmico:
@ajnavarro we simply can't know right now. This is why this is behind a flag. This is a "here be dragons, use at your own risk" kind of feature for cases where regular inner join is too slow and you have lots of memory to use. I don't see this enabled by default until we have a cost-based optimized (which is not likely in the foreseeable future).

ajnavarro:
we don't need a full cost-based optimizer for this. Maybe providing a method to tables to get an estimated size we can guess if it is worth it to fetch on memory or not. But we can discuss this implementation later.

erizocosmico:
@ajnavarro then this would only be applied to small tables. If a table is big (even if the end result of the branch after filters and so on is not), it will never be in memory.

ajnavarro:
then, maybe would be better allow to the user to activate or deactivate this functionality per query, using some session variable to do it (I would keep the actual env variable to globally activate the functionality anyways), WDYT?

erizocosmico:
@ajnavarro sounds good to me

erizocosmico:
Updated with the session variable

PR #609, opened 2019-01-29
sql/analyzer: fix use of declared aliases in the same projection

### LINKED ISSUE #590: alias used on a expression in the same projection should fail always
the following query:
```
select 
    language(blob_content, file_path) as lang, 
    substring(lang,1,1) as result
from refs 
NATURAL JOIN commits 
NATURAL JOIN commit_files 
NATURAL JOIN files 
where lang != '' limit 100;
```
Is returning results, but it should fail with the following error:
`ERROR 1105 (HY000): unknown error: column "lang" does not exist in scope, but there is an alias defined in this projection with that name. Aliases cannot be used in the same projection they're defined in`

mcarmonaa:
Fixes #590 

Since this is just a SQL syntax issue finally I've just added a new rule to check the aliases usage at the beginning of the analysis phase that runs once before the default rules. This way it doesn't interfere with the transformations made by the rest of the rules.

### DIFF
--- a/sql/analyzer/rules.go
+++ b/sql/analyzer/rules.go
@@ -27,6 +27,7 @@ var DefaultRules = []Rule{
 var OnceBeforeDefault = []Rule{
 	{"resolve_subqueries", resolveSubqueries},
 	{"resolve_tables", resolveTables},
+	{"check_aliases", checkAliases},
 }
 
 // OnceAfterDefault contains the rules to be applied just once after the
--- a/sql/analyzer/resolve_columns.go
+++ b/sql/analyzer/resolve_columns.go
@@ -12,6 +12,79 @@ import (
 	"gopkg.in/src-d/go-vitess.v1/vt/sqlparser"
 )
 
+func checkAliases(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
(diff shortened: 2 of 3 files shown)


sql/analyzer/rules.go
```diff
@@ -27,6 +27,7 @@ var DefaultRules = []Rule{
 var OnceBeforeDefault = []Rule{
 	{"resolve_subqueries", resolveSubqueries},
 	{"resolve_tables", resolveTables},
+	{"check_aliases", checkAliases},
```
erizocosmico: this should be a validation rule

mcarmonaa: validation rules are executed after regular rules, this must executed with the parsed plan before transformations

erizocosmico: But it's a validation rule. Why can't it be run in validation phase?

mcarmonaa: Since validation phase runs after the default rules, the plan to be validated could be modified enough to not keep this alias case.

An alias defined and used in the same projection in the original query, it couldn't be there in the plan after the transformations (reorder/remove/pushdown/joins...) (that's the case of the issue).

To be sure that a query which is making this is not allowed, the check must be done on the parsed plan or on a plan near to it.

PR #638, opened 2019-03-15
plan: compute all inner joins in memory if they fit

### LINKED ISSUE #577: Use in-memory join when one of the tables is small enough
This is just a future idea that I thought of leaving here for the future, as I was researching how to improve perf of the joins.

If we have two sides on a join we could try to use an in-memory cache for the smaller side instead of iterating it over and over. Depending on how expensive the underlying implementation is of iterating all over the rows, this might be a big performance boost.

What we need for this:
- Be able to estimate the size of each side (which will be more accurate the less filters and so on you have) if it's possible (inner join of an inner join might be tricky, for example).
- Optimisation rule that checks the size of each inner join side and if it can, replace it with an inner join that keeps the smaller side in memory instead of iterating it again and again.

Caveats:
- Estimating the size will be real hard and it's not portable, each impl must provide their own.
- Hard to decide a limit set in stone that's sensible for the in-memory join.

Sadly, this does not solve #454, because it can't and shouldn't be applied in every case. I've been digging on how to make a join not iterate one side N times, but it's the only way to do it without having a whole side in memory, which is not viable for large datasets.

erizocosmico:
Fixes #577

Because we do not have a way to estimate the cost of each side of
a join, it is really difficult to know when we can compute one in
memory. But not doing so, causes inner joins to be painfully slow,
as one of the branches is iterated multiple times.

This PR addresses this by ensuring that if the right branch of the
inner join fits in memory, it will be computed in memory even if
the in-memory mode has not been activated by the user.

An user can set the maximum threshold of memory the gitbase server
can use before considering the joins should not be performed in
memory using the `MAX_MEMORY_INNER_JOIN` environment variable or
the `max_memory_joins` session variable specifying the number of
bytes. The default value for this is the half of the available
physical memory on the operating system.

Because previously we had two iterators: `innerJoinIter` and
`innerJoinMemoryIter`, and now `innerJoinIter` must be able to do
the join in memory, `innerJoinMemoryIter` has been removed and
`innerJoinIter` replaced with a version that can work with three
modes:
- `unknownMode` we don't know yet how to perform the join, so keep
iterating until we can find out. By the end of the first full pass
over the right branch `unknownMode` will either switch to
`multipassMode` or `memoryMode`.
- `memoryMode` which computes the rest of the join in memory. The
iterator can have this mode before starting iterating if the user
activated the in memory join via session or environment vars, in
which case it will load all the right side on memory before doing
any further iteration. Instead, if the iterator started in
`unknownMode` and switched to this mode, it's guaranteed to already
have loaded all the right side. From that point on, they work in
exactly the same way.
- `multipassMode`, which was the previous default mode. Iterate the
right side of the join for each row in the left side. More expensive,
but less memory consuming. The iterator can not start in this mode,
and can only be switched to it from `unknownMode` in case the
memory used by the gitbase server exceeds the maximum amount of memory
either set by the user or by default.

### DIFF
--- a/sql/plan/innerjoin.go
+++ b/sql/plan/innerjoin.go
@@ -4,15 +4,52 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"runtime"
+	"strconv"
+	"strings"
 
 	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/pbnjay/memory"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 )
 
-const experimentalInMemoryJoinKey = "EXPERIMENTAL_IN_MEMORY_JOIN"
-const inMemoryJoinSessionVar = "inmemory_joins"
+const (
+	experimentalInMemoryJoinKey = "EXPERIMENTAL_IN_MEMORY_JOIN"
+	maxMemoryJoinKey            = "MAX_MEMORY_INNER_JOIN"
+	inMemoryJoinSessionVar      = "inmemory_joins"
+	memoryThresholdSessionVar   = "max_memory_joins"
+)
 
-var useInMemoryJoins = os.Getenv(experimentalInMemoryJoinKey) != ""
+var (
+	useInMemoryJoins = shouldUseMemoryJoinsByEnv()
+	// One fifth of the total physical memory available on the OS (ignoring the
+	// memory used by other processes).
+	defaultMemoryThreshold = memory.TotalMemory() / 5
+	// Maximum amount of memory the gitbase server can have in use before
+	// considering all inner joins should be done using multipass mode.
+	maxMemoryJoin = loadMemoryThreshold()
+)
+
+func shouldUseMemoryJoinsByEnv() bool {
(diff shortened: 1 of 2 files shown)


sql/plan/innerjoin.go
```diff
@@ -4,15 +4,46 @@ import (
...
+	maxMemoryJoinKey            = "MAX_MEMORY_INNER_JOIN"
+	inMemoryJoinSessionVar      = "inmemory_joins"
+	memoryThresholdSessionVar   = "max_memory_joins"
+)
+
+var (
+	useInMemoryJoins = os.Getenv(experimentalInMemoryJoinKey) != ""
+	// Half of the total physical memory available on the OS (ignoring the
+	// memory used by other processes).
+	defaultMemoryThreshold = memory.TotalMemory() / 2
```
ajnavarro: This is maybe too much? should reduce this to use just like 15-20% of the memory available? Take into account that the object's cache is only using 512mb as default param.

erizocosmico: Ok, I can change it to total memory / 5 or memory / 6

erizocosmico:
Fixed

erizocosmico:
Shall we merge this? @ajnavarro

PR #649, opened 2019-04-02
Added MySQL SLEEP function

### LINKED ISSUE #647: Request: SLEEP function
In MySQL you can run `SELECT SLEEP(10);`, and I think it would be good to have support for that in gitbase.

It would be very useful for development, and for testing, to have a way to simulate queries that take a long time. Specially for the code dealing with query cancellations.

https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_sleep

juanjux:
Adds MySQL SLEEP function, that seems to be useful for some kinds of testing.

Fixes #647.


sql/expression/function/sleep.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, nil
+	}
+
+	child, err = sql.Int32.Convert(child)
+	if err != nil {
+		return nil, err
+	}
+
+	if child.(int32) > 0 {
+		time.Sleep(time.Duration(child.(int32)) * time.Second)
```
kuba--: Based on mysql docs a fraction can be used, so my understanding is that someone can write `sleep(0.5)` to sleep for 500 ms. So, I think it's better to make it a float function.

juanjux: True! I'll fix it in the next push.

juanjux: Fixed and pushed, PTAL when you can.

kuba--:
```
go mod tidy
```
or rebuild:
```
/home/travis/gopath/pkg/mod/cache/vcs/55179c5d8c4db2eaed9fae4682d4c84a1fd3612df666b372bef3bbb997c9601f: exit status 128:
	fatal: unable to access 'https://go.googlesource.com/sync/': The requested URL returned error: 502
go: golang.org/x/sync@v0.0.0-20180314180146-1d60e4601c6f: unknown revision 1d60e4601c6f
go: error loading module requirements
```

erizocosmico:
@juanjux why is there a "foo" commit in the history? Leftover? :P

juanjux:
@erizocosmico :man_facepalming: yes, I call foo the commits I do with the intention of rebasing so I don't forget to do it. Clearly it didn't work this time.

juanjux:
@ajnavarro done!

juanjux:
PR rebased to the new `~master~.

ajnavarro:
@juanjux Check PR content, should we remove base64 Functions from this PR?

juanjux:
@ajnavarro done.

PR #667, opened 2019-04-10
sql: add support for intervals, DATE_SUB and DATE_ADD

### LINKED ISSUE #663: DATE_SUB and DATE_ADD functions
See https://github.com/src-d/gitbase/issues/784

erizocosmico:
Fixes #663

This PR introduces a few new features:
- Correctly parsing interval expressions.
- Allow using the + and - operators to subtract from and add to
  dates. e.g. `'2019-04-10' - INTERVAL 1 DAY`.
- New `DATE_ADD` function, which is essentially the same as
  `DATE + INTERVAL`.
- New `DATE_SUB` function, which is essentially the same as
  `DATE - INTERVAL`.
- Validation rule to ensure intervals are only used in certain
  specific places, such as DATE_SUB, DATE_ADD, + and -. Using it
  anywhere else is not valid SQL, but vitess does not catch those
  errors. Plus, even if it's an expression for convenience, its
  `Eval` method is a stub and panics, so it should not be used
  unless it's in an expression that knows how to deal with intervals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ We support and actively test against certain third-party clients to ensure compa
 |`CONCAT_WS(sep, ...)`|Concatenate any group of fields into a single string. The first argument is the separator for the rest of the arguments. The separator is added between the strings to be concatenated. The separator can be a string, as can the rest of the arguments. If the separator is NULL, the result is NULL.|
 |`CONNECTION_ID()`|Return the current connection ID.|
 |`COUNT(expr)`| Returns a count of the number of non-NULL values of expr in the rows retrieved by a SELECT statement.|
+|`DATE_ADD(date, interval)`|Adds the interval to the given date.|
+|`DATE_SUB(date, interval)`|Subtracts the interval from the given date.|
 |`DAY(date)`|Returns the day of the given date.|
 |`DAYOFWEEK(date)`|Returns the day of the week of the given date.|
 |`DAYOFYEAR(date)`|Returns the day of the year of the given date.|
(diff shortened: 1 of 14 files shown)


mcuadros:
I confirm that works properly with superset

erizocosmico:
Rebased

erizocosmico:
Rebased again

README.md
```diff
@@ -70,7 +70,8 @@ We support and actively test against certain third-party clients to ensure compa
 |`COUNT(expr)`| Returns a count of the number of non-NULL values of expr in the rows retrieved by a SELECT statement.|
 |`DATE_ADD(date, interval)`|Adds the interval to the given date.|
 |`DATE_SUB(date, interval)`|Subtracts the interval from the given date.|
-|`DAY(date)`|Returns the day of the given date.|
+|`DAY(date)`|Synonym for DAYOFMONTH().|
+|`DAYOFMONTH(date)`|Return the day of the month (0-31).|
```
kuba--: And I suppose we also need to add it to SUPPORT.md

erizocosmico:
Shall we merge this?

ajnavarro:
@erizocosmico I think changes on SUPPORTED.md file are missed.

erizocosmico:
@ajnavarro the one for DAYOFMONTH? That's not from this PR. DATE_ADD and DATE_SUB are already on SUPPORTED.md

README.md
erizocosmico: This is not from this PR

PR #672, opened 2019-04-11
Fix validation rule to detect tuples in projections or groupbys

juanjux:
Fixes src-d/gitbase/issues/785

Modifies a rule to check if distinct have a tuple argument (`DISTINCT(1,2)`), recursively. Unrelatedly, it also fix ordering in the README for the FROM_BASE64 function (my fault).

There are also a couple unrelated files with `go fmt` changes which I decided to add to this PR since it won't hurt.

### DIFF
--- a/sql/analyzer/optimization_rules.go
+++ b/sql/analyzer/optimization_rules.go
@@ -33,7 +33,6 @@ func optimizeDistinct(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, e
 	defer span.Finish()
 
 	a.Log("optimize distinct, node of type: %T", node)
-
 	if node, ok := node.(*plan.Distinct); ok {
 		var isSorted bool
 		_, _ = node.TransformUp(func(node sql.Node) (sql.Node, error) {
--- a/sql/analyzer/optimization_rules.go
+++ b/sql/analyzer/optimization_rules.go
@@ -1,7 +1,7 @@
 package analyzer
 
 import (
-	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-errors.v1"
 	"gopkg.in/src-d/go-mysql-server.v0/sql"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
 	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
@@ -33,6 +33,7 @@ func optimizeDistinct(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, e
 	defer span.Finish()
 
 	a.Log("optimize distinct, node of type: %T", node)
+
 	if node, ok := node.(*plan.Distinct); ok {
 		var isSorted bool
 		_, _ = node.TransformUp(func(node sql.Node) (sql.Node, error) {
--- a/sql/analyzer/resolve_columns.go
+++ b/sql/analyzer/resolve_columns.go
@@ -38,6 +38,31 @@ func checkAliases(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 	return n, err
 }
 
+func checkDistinctNoTuples(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
+	span, _ := ctx.Span("no_distinct_tuples")
+	defer span.Finish()
+
+	a.Log("check no tuples as distinct projection, node of type: %T", node)
+	var err error
+	if node, ok := node.(*plan.Distinct); ok {
+		_, err = node.TransformUp(func(node sql.Node) (sql.Node, error) {
+			project, ok := node.(*plan.Project)
+			if ok {
+				for _, col := range project.Projections {
+					_, ok := col.(expression.Tuple)
+					if ok {
+						return node, ErrDistinctTuple.New()
+					}
+				}
+			}
+
+			return node, nil
+		})
+	}
+
+	return node, err
+}
+
 func lookForAliasDeclarations(node sql.Expressioner) map[string]struct{} {
 	var (
 		aliases = map[string]struct{}{}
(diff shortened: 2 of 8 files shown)


sql/analyzer/optimization_rules.go
```diff
@@ -1,7 +1,7 @@
 package analyzer
 
 import (
-	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-errors.v1"
```
kuba--: Actually changes in this file you can discard.

sql/analyzer/validation_rules.go
```diff
@@ -3,7 +3,7 @@ package analyzer
 import (
 	"strings"
 
-	errors "gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-errors.v1"
```
kuba--: changes in this file we can also  discard

sql/analyzer/resolve_columns.go
```diff
@@ -38,6 +38,31 @@ func checkAliases(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 	return n, err
 }
 
+func checkDistinctNoTuples(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
```
kuba--: I don't know if it makes sense here, but did you test with wrapped nodes or subqueries?
Maybe we need to inspect distinct nodes recursively here, e.g. 
https://github.com/src-d/go-mysql-server/blob/master/sql/analyzer/assign_indexes.go#L43

erizocosmico:
It should not only be distinct, but anything that is projected. A tuple simply can't be projected (neither in Project nor GroupBy)

sql/analyzer/optimization_rules.go
juanjux: It was the result of gofmt over those files which doesn't hurt, but I'll discard to keep the PR tidier.

erizocosmico: This is a change goimports usually does. Instead of having to manually revert it every time I'd just keep it.

juanjux:
@erizocosmico change to detect tuples in projections and groupbys implemented, PTAL when you can.

sql/analyzer/resolve_columns.go
```diff
@@ -38,6 +38,41 @@ func checkAliases(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 	return n, err
 }
 
+func checkNoTuplesProjected(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
+	span, _ := ctx.Span("no_tuples_projected")
+	defer span.Finish()
+
+	a.Log("check no tuples as in projection, node of type: %T", node)
+	return node.TransformUp(func(node sql.Node) (sql.Node, error) {
+		project, ok := node.(*plan.Project)
```
erizocosmico: Project and GroupBy share a common interface: sql.Expressioner, which returns all expressions in a node.
You could leverage that common interface to reduce the code of this function to the following:

```go
switch node := node.(type) {
case *plan.Project, *plan.GroupBy:
        for _, e := range node.(sql.Expressioner).Expressions() {
                if sql.IsTuple(e.Type()) {
                        return nil, ErrTupleProjected.New()
                }
        }
}
```

Note that in this snippet I added a fix to your code. Not always you will have an `expression.Tuple` at the top level of an expression. The safe way to check if an expression is a tuple is if its type is a tuple. For this, we have a helper function named `sql.IsTuple`, so you can do `sql.IsTuple(expr.Type())`.

sql/analyzer/resolve_columns_test.go
```diff
@@ -75,6 +75,64 @@ func TestMisusedAlias(t *testing.T) {
...
+func TestNoTuplesProjected(t *testing.T) {
+	require := require.New(t)
+	f := getRule("no_tuples_projected")
+
+	table := mem.NewTable("mytable", sql.Schema{
+		{Name: "i", Type: sql.Int32},
+	})
+
+	node := plan.NewProject([]sql.Expression{
+		expression.NewTuple(
```
erizocosmico: We should add a new test case in which this `expression.NewTuple` is wrapped in a `expression.NewAlias`, for example. With that case, the current implementation would fail, but I already suggested a fix to that in my previous comment.

sql/analyzer/rules.go
```diff
@@ -66,4 +67,5 @@ var (
 	// ErrMisusedAlias is returned when a alias is defined and used in the same projection.
 	ErrMisusedAlias = errors.NewKind("column %q does not exist in scope, but there is an alias defined in" +
 		" this projection with that name. Aliases cannot be used in the same projection they're defined in")
+	ErrTupleProjected = errors.NewKind("unexpected tuple found, maybe remove the ()?")
```
erizocosmico: I would change the message to `unexpected tuple found in SELECT clause, tuples cannot be selected`. Because not always a tuple will be a tuple only as a result of having ().

juanjux: :+1:

juanjux:
Update: there was already a validation rule checking that there are not tuples in Projections or GroupBys (https://github.com/src-d/go-mysql-server/blob/master/sql/analyzer/validation_rules.go#L198), it even have tests with tuples (https://github.com/src-d/go-mysql-server/blob/master/sql/analyzer/validation_rules_test.go#L234) but adding a test with a DISTINCT->Projection->Tuple doesn't detect the error (and it crash like the issue reported if you do it on the REPL), so I'm not investigating why this existing rule doesn't work for this case.

juanjux:
Yes, with a recursive inspection the existing rule works beautifully, PTAL again since I removed basically everything from the previous commits and rewrote again.

sql/analyzer/resolve_columns.go
juanjux: Done (in the existing rule that wasn't doing it that way).

sql/analyzer/resolve_columns_test.go
juanjux: Done (in the testcase for the existing rule).

PR #682, opened 2019-04-17
sql/expression: relaxed json_extract

mcuadros:
In the current implementation a query stops to be executed when one of the rows is empty or doesn't fit the json path expression, this PR allows any failure on the query to a value, but fails on a wrong expression.

### DIFF
--- a/sql/expression/function/json_extract.go
+++ b/sql/expression/function/json_extract.go
@@ -69,10 +69,12 @@ func (j *JSONExtract) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 			return nil, err
 		}
 
-		result[i], err = jsonpath.JsonPathLookup(doc, path.(string))
+		c, err := jsonpath.Compile(path.(string))
 		if err != nil {
 			return nil, err
 		}
+
+		result[i], _ = c.Lookup(doc) // err ignored
 	}
 
 	if len(result) == 1 {
--- a/sql/expression/function/json_extract_test.go
+++ b/sql/expression/function/json_extract_test.go
@@ -1,6 +1,7 @@
 package function
 
 import (
+	"errors"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


sql/expression/function/json_extract.go
```diff
@@ -69,10 +69,12 @@ func (j *JSONExtract) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 			return nil, err
 		}
 
-		result[i], err = jsonpath.JsonPathLookup(doc, path.(string))
+		c, err := jsonpath.Compile(path.(string))
 		if err != nil {
 			return nil, err
 		}
+
+		result[i], _ = c.Lookup(doc) // err ignored
```
ajnavarro: Maybe it's worth it to give some info to the user in this case using `Warn`:

```go
ctx.Warn(0, err.Error())
```

sql/expression/function/json_extract_test.go
```diff
@@ -47,6 +47,8 @@ func TestJSONExtract(t *testing.T) {
 		row      sql.Row
 		expected interface{}
 	}{
+		{f2, sql.Row{nil, "$.b.c"}, nil},
+		{f2, sql.Row{json, "$.foo"}, nil},
```
ajnavarro: Maybe we can add `{f2, sql.Row{json, "FOO"}, nil},` too

sql/expression/function/json_extract.go
mcuadros: I believe that will be noise, because can be very common, since for example in my `LOC` if the language is not supported will be empty and then a warm

sql/expression/function/json_extract_test.go
mcuadros: This is going to be a error, I am adding it as it

PR #685, opened 2019-04-19
Suggest similar table/column/indexes names on missing errors

### LINKED ISSUE #684: [FEATURE] Suggest similar table/column names when one is not found in a query
Similarly to what some compilers do when the programmer mistype a symbol, the "table/column" not found could suggest alternative names from existing tables or columns in the specified table(s).

It could use levenshtein distance or some other text distance algorithm.

juanjux:
Fixes #684 

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -95,6 +95,11 @@ There are two authentication methods:
 - **None:** no authentication needed.
 - **Native:** authentication performed with user and password. Read, write or all permissions can be specified for those users. It can also be configured using a JSON file.
 
+## `internal/text_distance`
+
+Contains a function to return the most similar name from an
+array to a given one using the Levenshtein distance algorithm. Used for suggestions on errors.
+
 ## `internal/regex`
 
 go-mysql-server has multiple regular expression engines, such as oniguruma and the standard Go regexp engine. In this package, a common interface for regular expression engines is defined.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -95,9 +95,9 @@ There are two authentication methods:
 - **None:** no authentication needed.
 - **Native:** authentication performed with user and password. Read, write or all permissions can be specified for those users. It can also be configured using a JSON file.
 
-## `internal/text_distance`
+## `internal/similartext`
 
-Contains a function to return the most similar name from an
+Contains a function to `Find` the most similar name from an
 array to a given one using the Levenshtein distance algorithm. Used for suggestions on errors.
 
 ## `internal/regex`
--- a/internal/text_distance/text_distance.go
+++ b/internal/text_distance/text_distance.go
@@ -1,0 +1,52 @@
+package text_distance
+
+import (
+	"github.com/texttheater/golang-levenshtein/levenshtein"
+	"reflect"
+)
+
+// FindSimilarName returns the best match by Levenshtein distance
+// for the src string among the specified names.
+func FindSimilarName(names []string, src string) string {
+	minDistance := -1
+	var bestMatch string
+	s := []rune(src)
+	for _, name := range names {
+		r := []rune(name)
+		dist := levenshtein.DistanceForStrings(r, s, levenshtein.DefaultOptions)
+		if dist == 0 {
+			// Perfect match, shouldn't happen if this is used for errors
+			return name
+		}
+
+		if minDistance == -1 || dist < minDistance {
+			minDistance = dist
+			bestMatch = name
+		}
+	}
+
+	return bestMatch
+}
+
+// FindSimilarNameFromMap does the same as FindSimilarName but taking a map instead
+// of a string array as first argument.
+func FindSimilarNameFromMap(names interface{}, src string) string {
+	rnames := reflect.ValueOf(names)
+	if rnames.Kind() != reflect.Map {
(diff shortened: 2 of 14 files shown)


kuba--:
First of all I thought we won't pull any new dependencies. Distance function is just a one simple function. For instance here is sift4 (less general, but should be faster than Levenstein):
```
// Sift4 - common version
// online algorithm to compute the distance between two strings in O(n)
// maxOffset is the number of characters to search for matching letters
// maxDistance is the distance at which the algorithm should stop computing the value and just exit (the strings are too different anyway)
function sift4(s1, s2, maxOffset, maxDistance) {
    if (!s1||!s1.length) {
        if (!s2) {
            return 0;
        }
        return s2.length;
    }

    if (!s2||!s2.length) {
        return s1.length;
    }

    var l1=s1.length;
    var l2=s2.length;

    var c1 = 0;  //cursor for string 1
    var c2 = 0;  //cursor for string 2
    var lcss = 0;  //largest common subsequence
    var local_cs = 0; //local common substring
    var trans = 0;  //number of transpositions ('ab' vs 'ba')
    var offset_arr=[];  //offset pair array, for computing the transpositions

    while ((c1 < l1) && (c2 < l2)) {
        if (s1.charAt(c1) == s2.charAt(c2)) {
            local_cs++;
            var isTrans=false;
            //see if current match is a transposition
            var i=0;
            while (i<offset_arr.length) {
                var ofs=offset_arr[i];
                if (c1<=ofs.c1 || c2 <= ofs.c2) {
                    // when two matches cross, the one considered a transposition is the one with the largest difference in offsets
                    isTrans=Math.abs(c2-c1)>=Math.abs(ofs.c2-ofs.c1);
                    if (isTrans)
                    {
                        trans++;
                    } else
                    {
                        if (!ofs.trans) {
                            ofs.trans=true;
                            trans++;
                        }
                    }
                    break;
                } else {
                    if (c1>ofs.c2 && c2>ofs.c1) {
                        offset_arr.splice(i,1);
                    } else {
                        i++;
                    }
                }
            }
            offset_arr.push({
                c1:c1,
                c2:c2,
                trans:isTrans
            });
        } else {
            lcss+=local_cs;
            local_cs=0;
            if (c1!=c2) {
                c1=c2=Math.min(c1,c2);  //using min allows the computation of transpositions
            }
            //if matching characters are found, remove 1 from both cursors (they get incremented at the end of the loop)
            //so that we can have only one code block handling matches 
            for (var i = 0; i < maxOffset && (c1+i<l1 || c2+i<l2); i++) {
                if ((c1 + i < l1) && (s1.charAt(c1 + i) == s2.charAt(c2))) {
                    c1+= i-1; 
                    c2--;
                    break;
                }
                if ((c2 + i < l2) && (s1.charAt(c1) == s2.charAt(c2 + i))) {
                    c1--;
                    c2+= i-1;
                    break;
                }
            }
        }
        c1++;
        c2++;
        if (maxDistance)
        {
            var temporaryDistance=Math.max(c1,c2)-lcss+trans;
            if (temporaryDistance>=maxDistance) return Math.round(temporaryDistance);
        }
        // this covers the case where the last match is on the last token in list, so that it can compute transpositions correctly
        if ((c1 >= l1) || (c2 >= l2)) {
            lcss+=local_cs;
            local_cs=0;
            c1=c2=Math.min(c1,c2);
        }
    }
    lcss+=local_cs;
    return Math.round(Math.max(l1,l2)- lcss +trans); //add the cost of transpositions to the final result
}
```

The second thought I had is `maxDistance` parameter. In other words if the distance is > 3 it doesn't make sense to suggest anything because most likely it's a garbage.
I would suggest just a minimum < 3 and if you have more values with the same distance suggest all of them. For instance someone thought about `commits` or `ref_commits`

juanjux:
@kuba-- you're right, as we spoke on Slack, this could be a pretty shallow dependency, used a simplified and working on strings version of the Levenshtein function instead.

juanjux:
@kuba-- also implemented your suggestion of returning 0 to many results depending on the distance.

juanjux:
Looks like the Mono test in the CI matrix is failing for some missing package in some repository.

erizocosmico:
Ah, yes, the C# thing. Ignore it and don't worry.

ARCHITECTURE.md
```diff
@@ -95,6 +95,11 @@ There are two authentication methods:
 - **None:** no authentication needed.
 - **Native:** authentication performed with user and password. Read, write or all permissions can be specified for those users. It can also be configured using a JSON file.
 
+## `internal/text_distance`
+
```
erizocosmico: no underscore in package names please

internal/text_distance/text_distance.go
```diff
@@ -0,0 +1,107 @@
+package text_distance
```
erizocosmico: no underscore in package names please, this would be `textdistance`, but since what the exposed function is `FindSimilarNames`, I would name this package `similartext` and rename `FindSimilarNames` to just `Find` so that it can be used like `similartext.Find(bla, bla)`

ARCHITECTURE.md
juanjux: Done!

internal/text_distance/text_distance.go
juanjux: Both things done (plus `FindFromMap`).

PR #689, opened 2019-04-23
sql: implement HAVING clause

### LINKED ISSUE #56: TPC-H: unsupported feature: HAVING

erizocosmico:
Closes #56

This PR adds support for the HAVING SQL clause, that allows
filtering rows and has support for aggregation functions.

For this, the following changes have been made:
- New `Having` node, which is essentially a Filter, but it is
  a different node for the purpose of differentiating between the
  two of them during the analysis phase.
- Having is now parsed.
- A new rule for resolving the Having node has been added. Because
  of the way aggregations are executed (only inside a GroupBy node)
  it is not possible to execute them in any other node that's not a
  GroupBy. For this reason, this rule pushes down any aggregation
  on a Having node to its child, which is the GroupBy node. If
  the same aggregation is already done on the GroupBy, nothing will
  be added and the aggregation on the Having node will be replaced
  with a reference to the result of the aggregation in the GroupBy.
  Because pushing new aggregations to the GroupBy changes the
  resulting schema, a Project node is added wrapping the Having
  node projecting only the columns the GroupBy was initially
  projecting.

### DIFF
--- a/sql/analyzer/resolve_having.go
+++ b/sql/analyzer/resolve_having.go
@@ -1,0 +1,151 @@
+package analyzer
+
+import (
+	"reflect"
+
+	"gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression/function/aggregation"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/plan"
+)
+
+func resolveHaving(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) {
+	return node.TransformUp(func(node sql.Node) (sql.Node, error) {
+		having, ok := node.(*plan.Having)
+		if !ok {
+			return node, nil
+		}
+
+		if !having.Resolved() {
+			return node, nil
+		}
+
+		// If there are no aggregations there is no need to check anything else
+		// and we can just leave the node as it is.
+		if !hasAggregations(having.Cond) {
+			return node, nil
+		}
+
+		groupBy, ok := having.Child.(*plan.GroupBy)
+		if !ok {
+			return nil, errHavingNeedsGroupBy.New()
+		}
+
+		var aggregate = make([]sql.Expression, len(groupBy.Aggregate))
+		copy(aggregate, groupBy.Aggregate)
+
+		// We need to find all the aggregations in the having that are already present in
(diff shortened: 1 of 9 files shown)


sql/analyzer/resolve_having.go
```diff
@@ -0,0 +1,154 @@
...
+		return result, nil
+	})
+}
+
+func aggregationEquals(a, b sql.Expression) bool {
+	if alias, ok := b.(*expression.Alias); ok {
+		return aggregationEquals(a, alias.Child)
+	}
+
+	if alias, ok := a.(*expression.Alias); ok {
```
juanjux: Maybe if both are aliases the children could be extracted directly here from both to avoid one function call, but this is super nitpicking.

sql/analyzer/resolve_having.go
```diff
@@ -0,0 +1,154 @@
...
+	case *aggregation.Count:
+		// it doesn't matter what's inside a Count, the result will be
+		// the same.
+		_, ok := b.(*aggregation.Count)
+		return ok
+	case *aggregation.Sum,
+		*aggregation.Avg,
+		*aggregation.Min,
+		*aggregation.Max:
+		fmt.Println(a, b, reflect.DeepEqual(a, b))
```
juanjux: Debug stuff?

sql/analyzer/resolve_having.go
```diff
@@ -0,0 +1,154 @@
...
+	if alias, ok := b.(*expression.Alias); ok {
+		return aggregationEquals(a, alias.Child)
+	}
+
+	if alias, ok := a.(*expression.Alias); ok {
+		return aggregationEquals(alias.Child, b)
+	}
+
+	switch a := a.(type) {
+	case *aggregation.Count:
```
kuba--: I also wonder about other functions, e.g. `SUBSTRING`.
What if you have:
```
SELECT str, COUNT(n) FROM table GROUP BY n HAVING SUBSTR(str, 0, 1) = 'a'
```

erizocosmico: Those work exacty as a Filter node

sql/analyzer/resolve_having.go
erizocosmico: yup

sql/analyzer/resolve_having.go
erizocosmico: done

erizocosmico:
Rebasing. Some tests are failing now after the rebase.

erizocosmico:
Rebased and regression fixed.

PR #694, opened 2019-04-24
sql: convert dates outside of supported time range to correct date

erizocosmico:
According to MySQL, dates should range between 1000-01-01 00:00:00
and 9999-12-31 23:59:59. This PR changes the behaviour of Convert
methods of Timestamp and Date types so that values returned by this
method are always within that range.

It also adds an exposed utility function named `ToSupportedTimeRange`
that converts a time to the supported time range for clients to use.

The readme now includes instructions for datasource implementors to
always return their dates within the supported range using the
provided utility function.
The reason go-mysql-server does not handle that for data sources by
default is because there is no obvious place to do so in a generic
way. Tables should be the ones returning correct values.


juanjux:
Why the datasource implementors must certainly provide the dates in the right format, I wonder if adding some more checks in our code for the dates to be in the right range (in the implementations of DATE, internal, etc) could be beneficial or have some performance problems.

erizocosmico:
I've been checking the behaviour of MySQL and it's a little bit more complex than this. Don't merge.

erizocosmico:
Closing this. Needs different solution, as explained in #698

PR #697, opened 2019-04-24
pilosa: set max map count to 0 to avoid index issues on windows

erizocosmico:


kuba--:
Travis is failing, because we need to take off:
https://github.com/src-d/go-mysql-server/blob/master/.travis.yml#L25

PR #700, opened 2019-05-03
add greatest and least

juanjux:
- Adds MySQL `GREATEST` and `LEAST` functions.
- Sort function names in `SUPPORTED.md`.


README.md
```diff
@@ -77,10 +77,12 @@ We support and actively test against certain third-party clients to ensure compa
 |`DAYOFYEAR(date)`|Returns the day of the year of the given date.|
 |`FLOOR(number)`|Return the largest integer value that is less than or equal to `number`.|
 |`FROM_BASE64(str)`|Decodes the base64-encoded string str.|
+|`GREATEST(...)`|Return the greatest numeric or string value.|
 |`HOUR(date)`|Returns the hours of the given date.|
```
kuba--: returns

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+		if sql.IsTuple(arg.Type()) {
+			return nil, sql.ErrInvalidType.New("tuple")
+		}
+	}
+
+	return &Greatest{args}, nil
+}
+
+// Type implements the Expression interface.
+func (f *Greatest) Type() sql.Type { return sql.Float64 }
```
kuba--: float? so what about strings?

kuba--: float?

juanjux: Looks like I left this when I started to develop this just for floats. But it works :-? Is there any way I can specify several types there? (this could return float64, int64 or string depending on the arguments).

README.md
juanjux: Fixed, thanks.

sql/expression/function/greatest_least.go
erizocosmico: Take a look at the Comparison type logic to determine which type to use.

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// Greatest returns the argument with the greatest numerical or string value. It allows for
+// numeric (ints anf floats) and string arguments and will return the used type
+// when all arguments are of the same type or floats if there are numerically
+// convertible strings or integers mixed with floats. When ints or floats
+// are mixed with non numerically convertible strings, those are ignored.
+type Greatest struct {
+	args []sql.Expression
```
erizocosmico: args should be exported, just so it can be transformed in a rule

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+var ErrUnsupportedType = errors.NewKind("unsupported type for greatest/latest argument: %T")
+
+// NewGreatest creates a new Greatest UDF
+func NewGreatest(args ...sql.Expression) (sql.Expression, error) {
+	if len(args) == 0 {
+		return nil, sql.ErrInvalidArgumentNumber.New("GREATEST", "1 or more", 0)
+	}
+
+	for _, arg := range args {
+		if !arg.Resolved() {
```
erizocosmico: When it's first parsed, nothing that's not a literal will be resolved, so they will be ignored. So this if should be removed.

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+// NewLeast creates a new Least UDF
+func NewLeast(args ...sql.Expression) (sql.Expression, error) {
+	if len(args) == 0 {
+		return nil, sql.ErrInvalidArgumentNumber.New("LEAST", "1 or more", 0)
+	}
+
+	for _, arg := range args {
+		// Don't perform this check until it's resolved. Otherwise we
+		// can't get the type for sure.
+		if !arg.Resolved() {
```
erizocosmico: same as in Greatest

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+	return greatestNum, nil
+}
+
+// Least returns the argument with the least numerical or string value. It allows for
+// numeric (ints anf floats) and string arguments and will return the used type
+// when all arguments are of the same type or floats if there are numerically
+// convertible strings or integers mixed with floats. When ints or floats
+// are mixed with non numerically convertible strings, those are ignored.
+type Least struct {
+	args []sql.Expression
```
erizocosmico: same as in Greatest

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,329 @@
...
+		if sql.IsTuple(arg.Type()) {
+			return nil, sql.ErrInvalidType.New("tuple")
+		}
+	}
+
+	return &Least{args}, nil
+}
+
+// Type implements the Expression interface.
+func (f *Least) Type() sql.Type { return sql.Float64 }
```
erizocosmico: same as in Greatest

engine_test.go
```diff
@@ -1036,6 +1036,14 @@ var queries = []struct {
 			{string("first row"), int64(1)},
 		},
 	},
+	{
+		`SELECT GREATEST(1, 2, 3, 4)`,
+		[]sql.Row{{int64(4)}},
+	},
+	{
+		`SELECT LEAST(1, 2, 3, 4)`,
```
ajnavarro: could you add some tests using strings too?

sql/expression/function/greatest_least_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+import (
+	"github.com/stretchr/testify/require"
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+	"gopkg.in/src-d/go-mysql-server.v0/sql/expression"
+	"testing"
+)
+
+func TestGreatest(t *testing.T) {
+	t.Run("null", func(t *testing.T) {
```
erizocosmico: Instead of repeating the same boilerplate over and over, what do you think about a table-driver test?

```go
func TestGreatest(t *testing.T) {
    testCases := []struct{
        name string
        args []sql.Expression
        expected interface{}
    }{
        {
            "null",
            []sql.Expression{
                expression.NewLiteral(nil, sql.Null),
			    expression.NewLiteral(5, sql.Int64),
			    expression.NewLiteral(1, sql.Int64),
            },
            nil,
        },
        {
            "negative and all ints",
            []sql.Expression{
                expression.NewLiteral(int64(-1), sql.Int64),
			    expression.NewLiteral(int64(5), sql.Int64),
			    expression.NewLiteral(int64(1), sql.Int64),
            },
            int64(5),
        },
        // ...
    }

    for _, tt := range testCases {
        t.Run(tt.name, func(t *testing.T) {
            require := require.New(t)

            f, err := NewGreatest(tt.args...)
            require.NoError(err)

            output, err := f.Eval(sql.NewEmptyContext(), nil)
            require.NoError(err)
            require.Equal(tt.expected, output)
        })
    }
}
```

sql/expression/function/greatest_least_test.go
```diff
@@ -0,0 +1,209 @@
...
+		f, err := NewGreatest(
+			expression.NewLiteral(int64(-1), sql.Int64),
+			expression.NewLiteral(int64(5), sql.Int64),
+			expression.NewLiteral(int64(1), sql.Int64),
+		)
+		require.NoError(err)
+
+		v, err := f.Eval(sql.NewEmptyContext(), nil)
+		require.NoError(err)
+		require.IsType(int64(1), v)
```
erizocosmico: this assert is redundant. require.Equal already fails if the type is not the same as the one of the expected value

sql/expression/function/greatest_least_test.go
```diff
@@ -0,0 +1,209 @@
...
+		require.NoError(err)
+
+		v, err := f.Eval(sql.NewEmptyContext(), nil)
+		require.NoError(err)
+		require.IsType("", v)
+		require.Equal("bbb", v)
+	})
+}
+
+func TestLeast(t *testing.T) {
```
erizocosmico: same as TestGreatest

engine_test.go
erizocosmico: And some tests using table columns

sql/expression/function/greatest_least.go
juanjux: Fixed in next push.

sql/expression/function/greatest_least.go
juanjux: Fixed in next push.

engine_test.go
juanjux: @ajnavarro there are already tests with strings (the ones called "all strings", "all strings and empty", and a couple mixing strings with other types).

sql/expression/function/greatest_least_test.go
juanjux: Fixed in next push.

sql/expression/function/greatest_least_test.go
juanjux: Ditto.

sql/expression/function/greatest_least.go
juanjux: :+1:

sql/expression/function/greatest_least.go
juanjux: Done.

sql/expression/function/greatest_least_test.go
juanjux: Done, thanks.

sql/expression/function/greatest_least.go
juanjux: `sql/expression/comparison.go::Comparison` seems to return a `Boolean`. I updated the PR adding a type member and updating it after `Eval`, but I'm not sure that's right, could `Type()` be called before `Eval()`?

erizocosmico: Yup, type can and will be called before Eval.

I meant this logic: https://github.com/src-d/gitbase/issues/808

juanjux: Implemented, thanks for the tip.

juanjux:
Type deduction implemented as suggested plus some factorizing for `NewXXX` and `Eval` to avoid duplicated code. Only missing the tests with a table and columns (currently checking learn how to do that first :)

engine_test.go
juanjux: @ajnavarro , @erizocosmico added tests with strings and tables to this file, PTAL.

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,309 @@
...
+	"gopkg.in/src-d/go-mysql-server.v0/sql"
+)
+
+// compEval is used to implement Greatest/Least Eval() using a comparison function
+func compEval(
+	returnType sql.Type,
+	args []sql.Expression,
+	ctx *sql.Context,
+	row sql.Row,
+	cmp compareFn) (interface{}, error) {
```
erizocosmico: s/cmp compareFn) (interface{}, error) {/cmp compareFn,
) (interface{}, error) {/

sql/expression/function/greatest_least_test.go
```diff
@@ -0,0 +1,183 @@
...
+            []sql.Expression{
+				expression.NewLiteral(string("10.5"), sql.Text),
+				expression.NewLiteral(string("foobar"), sql.Int64),
+				expression.NewLiteral(int64(5), sql.Int64),
+				expression.NewLiteral(int64(1), sql.Int64),
+			},
+			float64(10.5),
+		},
+		{
+            "float mixed",
```
erizocosmico: is this gofmt'd? The indentation seems off

sql/expression/function/greatest_least.go
juanjux: Done.

sql/expression/function/greatest_least_test.go
juanjux: Ups, fixed.

sql/expression/function/greatest_least.go
```diff
@@ -0,0 +1,310 @@
...
+	var selectedString string
+
+	for i, arg := range args {
+		val, err := arg.Eval(ctx, row)
+		if err != nil {
+			return nil, err
+		}
+
+		switch t := val.(type) {
+		case int, int32, int64:
```
kuba--: What about `unsigned` variants?
I don't want to be too much `tiquismiquis`, but maybe we should stick just to 2 options - `string` or `float64`.
I remember we had the same _pleasure_ fighting with different types for arithmetic functions (max, min, avg, ...) and ultimately we decided to stick to float64.

juanjux: Well, there is not much difference between supporting only `string` and `float64` and also supporting `int/32/64` but not unsigned.

kuba--: I just meant:
```
if string {
// convert to string
} else {
// convert to float64
}
```

if we wanna keep the current implementation it's fine, if we add int8, int16 and unsigned variants

juanjux: @kuba [PTAL](https://github.com/src-d/go-mysql-server/pull/700/commits/7cc58ce9b8e84cd84299e1946b5d77f0934a1899)

ajnavarro:
@juanjux LGTM after resolving conflicts!

juanjux:
@ajnavarro done!

PR #701, opened 2019-05-06
sql/analyzer: alias projected columns wrapped in converts

erizocosmico:
Fixes required to work on gitbase. It was caught there because there are way more complex queries in the gitbase test suite.

Still needs some fixes to work on gitbase, because squash rule works a bit weirdly with this.


erizocosmico:
Don't merge, just found a test case that makes this fail.

erizocosmico:
All working now

PR #703, opened 2019-05-07
sql/analyzer: run convert_dates rule after prune_columns

erizocosmico:
After a while debugging on gitbase I realised only this fix would also fix all the problems it had with squash rule with much less code.

Analyzer bugs are weird 🤷‍♂


juanjux:
I wonder the debug process you followed to discover that, anyway you're a coding monster.

erizocosmico:
@juanjux went down the rabbit hole adding more logic to the squash rule until I saw prune_columns was not working as expected because of things convert_dates was doing. Executing it later made all problems go away. Actually, the only requirement convert_dates has is to execute before pushdown and after everything has been resolved. So I moved it just before pushdown.

erizocosmico:
Shall we merge this @ajnavarro?

PR #704, opened 2019-05-07
sql: use correct MySQL display name for types

### LINKED ISSUE #702: Incorrect type name for int64 in create table
More info: https://github.com/src-d/gitbase/issues/808

There might be more types with a wrong name in CREATE TABLE. I'd check all uint/int types and floats to see if it's the case.
Also, there might be more affected statements other than CREATE TABLE if they output in any way the type of the columns.

erizocosmico:
Fixes #702


erizocosmico:
Fixed

erizocosmico:
Shall we merge this @ajnavarro?

PR #708, opened 2019-05-13
sql/analyzer: do not use plan.Inspect to find available tables

erizocosmico:
Friday night this possible bug came to my mind, but PR was merged because I got a chance to fix it tomorrow 😅 

Because of how plan.Inspect works, whenever a table is found in a branch the other branches are not inspected anymore. Instead, doing recursive calls all branches are visited.

### DIFF
--- a/sql/analyzer/resolve_columns.go
+++ b/sql/analyzer/resolve_columns.go
@@ -239,27 +239,31 @@ func getColumnsInNodes(nodes []sql.Node, columns map[string][]string) {
 }
 
 func getNodeAvailableTables(n sql.Node) map[string]string {
-	var tables = make(map[string]string)
-	for _, c := range n.Children() {
-		plan.Inspect(c, func(n sql.Node) bool {
-			switch n := n.(type) {
-			case *plan.SubqueryAlias, *plan.ResolvedTable:
-				name := strings.ToLower(n.(sql.Nameable).Name())
-				tables[name] = name
-				return false
-			case *plan.TableAlias:
-				switch t := n.Child.(type) {
-				case *plan.ResolvedTable, *plan.UnresolvedTable:
-					name := strings.ToLower(t.(sql.Nameable).Name())
-					alias := strings.ToLower(n.Name())
-					tables[alias] = name
-				}
-			}
-
-			return true
-		})
-	}
+	tables := make(map[string]string)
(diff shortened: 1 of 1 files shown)


sql/analyzer/resolve_columns.go
```diff
@@ -239,27 +239,31 @@ func getColumnsInNodes(nodes []sql.Node, columns map[string][]string) {
...
-		plan.Inspect(c, func(n sql.Node) bool {
-			switch n := n.(type) {
-			case *plan.SubqueryAlias, *plan.ResolvedTable:
-				name := strings.ToLower(n.(sql.Nameable).Name())
+	tables := make(map[string]string)
+	getNodesAvailableTables(tables, n.Children()...)
+	return tables
+}
+
+func getNodesAvailableTables(tables map[string]string, nodes ...sql.Node) {
```
kuba--: maybe it can also return a tables map - WDYT?

erizocosmico: The reason why this does not return a table map is because it takes one as a parameter to add stuff to it and avoid having to merge maps since this is called recursively. It could return a map but it's pointless because it will return the same map as the one it receives.

sql/analyzer/resolve_columns.go
```diff
@@ -239,27 +239,31 @@ func getColumnsInNodes(nodes []sql.Node, columns map[string][]string) {
...
+		case *plan.SubqueryAlias, *plan.ResolvedTable:
+			name := strings.ToLower(n.(sql.Nameable).Name())
+			tables[name] = name
+		case *plan.TableAlias:
+			switch t := n.Child.(type) {
+			case *plan.ResolvedTable, *plan.UnresolvedTable:
+				name := strings.ToLower(t.(sql.Nameable).Name())
+				alias := strings.ToLower(n.Name())
+				tables[alias] = name
+				// Also add the ame of the table because you can refer to a
```
kuba--: Also add the `name`...

sql/analyzer/resolve_columns.go
kuba--: Makes sense. It's still tail recursion, so OK.

erizocosmico:
Shall we merge this, @ajnavarro?

PR #712, opened 2019-05-15
sql: implement LEFT and RIGHT join

### LINKED ISSUE #707: LEFT OUTER JOIN support
LEFT OUTER JOIN support would be nice.

For gitbase, this might be used to check repositories where a given file is absent. Note that for the use cases I'm trying, getting the LEFT OUTER JOIN resolved in Spark with gitbase-spark-connector would be just fine.

erizocosmico:
Closes #707

This PR implements LEFT and RIGHT join. Initial issue only required
LEFT join, but the effort to also add support to RIGHT join was
minimal, so it was added as well.

The join iterator, previously inner join iterator, has been generalized
so it can work for all joins and share all the logic in a single
component instead of having three iterators handling the complex logic
of how to compute the joins (in memory, multipass, etc).

The nodes, however, have very subtle differences that make it worthless
to abstract because most methods need to be different. The only abstractable
part was the RowIter method, which was extracted to a function that's
called from InnerJoin, LeftJoin and RightJoin nodes.


erizocosmico:
CI fails because of C# integration test

ajnavarro:
could you rebase please?

erizocosmico:
Rebased

PR #713, opened 2019-05-15
server: guard sessions access in mutex

erizocosmico:
Could panic with "concurrent map write" error


erizocosmico:
CI fails because of C# integration test.

PR #720, opened 2019-05-22
sql: implement EXPLODE and generators

erizocosmico:
This pull request implements the EXPLODE expressions and generators,
which are used together to provide generation of rows based on a
column or expression that can yield multiple values.

For this, there have been some quite core changes:
- SQL method of sql.Type now returns the sqltypes.Value and an error.
  This avoids panics in this method. Since generators can yield errors
  it's better to just return an error and not just kill the server.
- Array type can now internally handle either arrays or generators,
  making it transparent for the user. If it's used in an EXPLODE
  expression, the generator will be used. Otherwise, it will be
  converted automatically to an array and used without the user even
  knowing what happened behind the scenes. That way, we don't need
  yet another type to represent generators.

Aside from that, there are some new additions:
- New EXPLODE function, which is just a placeholder. EXPLODE type is the
  underlying type of its argument for resolution purposes. During analysis
  Explode nodes will be replaced by Generate functions and a Generate node.
- New Generate function, which is the non-placeholder version of EXPLODE
  and the one that goes into the final execution tree. This one returns
  the same type as its argument and let's the Generate plan node be the
  one that returns the underlying type once the values are generated.
- Generate node, which wraps a Project in which there is one and only one
  explode expression.
- resolve_generators analysis rule, which will turn projects with an
  Explode expression into a Generate node with the project as a children,
  replacing the Explode expressions with Generate expressions.
- validate_explode_usage validation rule, which will ensure explode is
  not used outside a Project node.

### DIFF
--- a/sql/analyzer/resolve_generators.go
+++ b/sql/analyzer/resolve_generators.go
@@ -1,0 +1,97 @@
+package analyzer
+
+import (
+	"gopkg.in/src-d/go-errors.v1"
+	"github.com/src-d/go-mysql-server/sql"
+	"github.com/src-d/go-mysql-server/sql/expression"
+	"github.com/src-d/go-mysql-server/sql/expression/function"
+	"github.com/src-d/go-mysql-server/sql/plan"
+)
+
+var (
+	errMultipleGenerators = errors.NewKind("there can't be more than 1 instance of EXPLODE in a SELECT")
+	errExplodeNotArray    = errors.NewKind("argument of type %q given to EXPLODE, expecting array")
+)
+
+func resolveGenerators(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
+	return n.TransformUp(func(n sql.Node) (sql.Node, error) {
+		p, ok := n.(*plan.Project)
+		if !ok {
+			return n, nil
+		}
+
+		projection := p.Projections
+
+		g, err := findGenerator(projection)
+		if err != nil {
+			return nil, err
+		}
+
+		// There might be no generator in the project, in that case we don't
+		// have to do anything.
+		if g == nil {
+			return n, nil
+		}
+
+		projection[g.idx] = g.expr
+
+		var name string
+		if n, ok := g.expr.(sql.Nameable); ok {
+			name = n.Name()
+		} else {
+			name = g.expr.String()
+		}
+
+		return plan.NewGenerate(
+			plan.NewProject(projection, p.Child),
+			expression.NewGetField(g.idx, g.expr.Type(), name, g.expr.IsNullable()),
+		), nil
+	})
+}
+
+type generator struct {
+	idx  int
+	expr sql.Expression
+}
+
+// findGenerator will find in the given projection a generator column. If there
+// is no generator, it will return nil.
+// If there are is than one generator or the argument to explode is not an
+// array it will fail.
+// All occurrences of Explode will be replaced with Generate.
+func findGenerator(exprs []sql.Expression) (*generator, error) {
+	var g = &generator{idx: -1}
+	for i, e := range exprs {
+		var found bool
+		switch e := e.(type) {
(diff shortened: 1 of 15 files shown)


sql/plan/generate.go
```diff
@@ -0,0 +1,152 @@
...
+	if !ok {
+		return nil, fmt.Errorf("column of Generate node transformed into %T, must be GetField", col)
+	}
+
+	return NewGenerate(child, field), nil
+}
+
+func (g *Generate) String() string {
+	tp := sql.NewTreePrinter()
+	tp.WriteNode("Generate(%s)", g.Column)
```
golangcibot: Error return value of `tp.WriteNode` is not checked (from `errcheck`)

sql/plan/generate.go
```diff
@@ -0,0 +1,152 @@
...
+	tp.WriteChildren(g.Child.String())
```
golangcibot: Error return value of `tp.WriteChildren` is not checked (from `errcheck`)

sql/analyzer/resolve_generators.go
```diff
@@ -0,0 +1,97 @@
...
+	expr sql.Expression
+}
+
+// findGenerator will find in the given projection a generator column. If there
+// is no generator, it will return nil.
+// If there are is than one generator or the argument to explode is not an
+// array it will fail.
+// All occurrences of Explode will be replaced with Generate.
+func findGenerator(exprs []sql.Expression) (*generator, error) {
+	var g = &generator{idx: -1}
```
kuba--: Maybe I'm wrong but we want to find the only one, first generator, so maybe we can move `found` outside the loop and get rid of idx.

erizocosmico: No, we want to find one (because there can only be one) but we also want to error if there are more generators.
If we move found outside you have no way of telling whether the found happened in this iteration or in previous ones and you don't know if you should error or not.

sql/generator.go
```diff
@@ -0,0 +1,57 @@
...
+	// Close the generator and dispose resources.
+	Close() error
+}
+
+// ErrNotGenerator is returned when the value cannot be converted to a
+// generator.
+var ErrNotGenerator = errors.NewKind("cannot convert value of type %T to a generator")
+
+// ToGenerator converts a value to a generator if possible.
+func ToGenerator(v interface{}) (Generator, error) {
```
juanjux: Since it only really converts arrays, could be called ArrayToGenerator.

erizocosmico: ArrayToGenerator suggests you need to pass an array and get a generator. This, however, receives any kind of value and returns a Generator or an error if the value could not be converted. The fact that arrays are converted is an implementation detail.

juanjux: It is expected to convert other types in the future? (like maps for example).

erizocosmico: Nope, because there are not any other types that can be a generator besides  arrays and `Generator`. But still calling it ArrayToGenerator implies the input is always an Array, which is not the case. It's meant to return the generator as is if the input is a generator, convert array to a generator and return it or fail in any other case.

juanjux: Mmm, then since the `interface{}` doesn't provide documentation about it to the user this could be specified in the function doc.

erizocosmico: The user does not need to know which types are handled here. Only needs to know that if the value can be converted to a generator it will, if not, it will error.

juanjux:
Needs rebase, can then be merged? (I can use this on the BLAME PR for the integration tests).

ajnavarro:
LGTM after rebase.

erizocosmico:
Rebased

ajnavarro:
@erizocosmico could you check tests?

erizocosmico:
I rebased, but didn't remember to change the imports to the new paths 😅 

Fixed

Issue #726, opened 2019-05-24
gitbase dies for query with WHERE NOT <not boolean value>


zurk:
If you try to query something like: 
```
select repository_id, commit_author_when, commit_parents from commits WHERE NOT commit_parents
```
gitbase dies. It is not a valid SQL query, because `commit_parents` type is not boolean, but I expect that `gitbase` remains available for new queries.

It is useful to have some general check if gitbase fails to process the query and do not shut down if the query was bad.

smola:
I think it's ok to just check for bool on evaluation if you introduce the right `CAST` during analysis.

PR #731, opened 2019-05-28
Make boolean expression compatible with MySQL

kuba--:
Fixes: https://github.com/src-d/go-mysql-server/issues/726


kuba--:
Every string is `false`. Number, Time, Duration are converted into `int` (float is round), `0` is `false`, otherwise `true`.

PR #734, opened 2019-05-29
Add metrics for queries (count parallel queries)

kuba--:

It's a first iteration to expose metrics for prometheus. So far, it contains:
- counters and histograms for queries (also number of parallel queries is counted)
- gauge and histograms for pilosa index.

If we find it useful we can add more either on mysql or gitbase level.

Related to https://github.com/src-d/gitbase/issues/815


engine.go
```diff
@@ -127,3 +149,15 @@ func (e *Engine) AddDatabase(db sql.Database) {
 func (e *Engine) Init() error {
 	return e.Catalog.LoadIndexes(e.Catalog.AllDatabases())
 }
+
+func logQuery(ctx *sql.Context, query string) func() {
+	logrus.WithField("query", query).Debug("executing query")
+	span, ctx := ctx.Span("query", opentracing.Tag{Key: "query", Value: query})
```
golangcibot: SA4006: this value of `ctx` is never used (from `staticcheck`)

PR #736, opened 2019-05-30
replace all gopkg.in imports and use only go mods

erizocosmico:
First step of moving gitbase to go modules is moving this to go modules. We already do the versioning with it, so there's no point in versioning with gopkg.in as well.


erizocosmico:
Shall we merge this, @ajnavarro?

PR #737, opened 2019-05-30
Fix parsing of quoted identifiers in SHOW CREATE TABLE queries

### LINKED ISSUE #733: Cannot parse "show create table `a.b`"
Queries like 

```sql
select * from `a.b`
```

parse, but

```sql
show create table `a.b`
```
does not. It fails with

```
expecting "`" but got "." instead
```

I think https://github.com/src-d/go-mysql-server/blob/1d000c3020dc730580c98f534113c6ba7528ccb5/sql/parse/util.go#L236-L240 is to blame because `readIdent` tries to parse a valid identifier, but since we are in backticks we should be accepting almost anything here instead.

andremarianiello:
Fixes #733

### DIFF
--- a/sql/parse/util.go
+++ b/sql/parse/util.go
@@ -202,7 +202,7 @@ func readIdent(ident *string) parseFunc {
 func readQuotedIdent(ident *string) parseFunc {
 	return func(r *bufio.Reader) error {
 		var buf bytes.Buffer
-		if err := readLetter(r, &buf); err != nil {
+		if err := readValidQuotedIdentRune(r, &buf); err != nil {
 			return err
 		}
 
--- a/sql/parse/util.go
+++ b/sql/parse/util.go
@@ -144,6 +144,36 @@ func readValidIdentRune(r *bufio.Reader, buf *bytes.Buffer) error {
 	return nil
 }
 
+func readValidQuotedIdentRune(r *bufio.Reader, buf *bytes.Buffer) error {
+	bs, err := r.Peek(2)
+	if err != nil {
+		return err
+	}
+
+	if bs[0] == '`' && bs[1] == '`' {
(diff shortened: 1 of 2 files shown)


sql/parse/util.go
```diff
@@ -169,6 +199,26 @@ func readIdent(ident *string) parseFunc {
 	}
 }
 
+func readQuotedIdent(ident *string) parseFunc {
+	return func(r *bufio.Reader) error {
+		var buf bytes.Buffer
+		if err := readLetter(r, &buf); err != nil {
```
erizocosmico: why is this necessary? Doesn't this force the first character to be a letter?

andremarianiello: I copied that from `readIdent`, but I didn't think about whether that is necessary. The name `readLetter` does imply that the first character must be a letter, but since the test cases which have identifiers starting with non-letters pass, this seems to not be the case. Perhaps it has another purpose? Regardless, I agree that `readLetter` isn't needed here. Will remove.

andremarianiello: Rather than removing completely, I replaced it with a call to `readValidQuotedIdentRune` so that the identifier cannot be empty.

erizocosmico:
@kuba-- in this case SHOW CREATE TABLE ´a.b´ should not be split. Anything inside backticks is an identifier and can't be split. The only reason to have them separated would be a.b or `a`.`b`.

kuba--:
if it's a case (that it should not be split) we can just add _a dot_ ('.'), here:
https://github.com/src-d/go-mysql-server/blob/master/sql/parse/util.go#L136
and problem looks as solved.

erizocosmico:
@kuba-- nope, because when the identifier is not quoted, the dot is not a valid rune for an identifier.

PR #739, opened 2019-06-03
sql/analyzer: back-propagate expression names after adding convert

erizocosmico:


ajnavarro:
Shall we avoid add date casts to fields that are already a Date?

erizocosmico:
Fields that are already date are exactly the reason why we added this rule. Data sources may return dates that are outside the valid range for MySQL.

PR #744, opened 2019-06-04
Add metrics (engine, analyzer, regex, pilosa) based on go-kit interface.

kuba--:
Orchestrate mysql server by metrics based on `go-kit` metrics interface.

Generic implementation (to satisfy prometheus, etc.) comes with some constraints (like you cannot reregister the same metric, you cannot reuse constructors, so we cannot pass builders to some kind of registry etc.). So, we ended up with following proposal, where you have to initialize global vars per package.

By default all metrics are based on `discard` implementation which is basically `noop`.


engine.go
```diff
@@ -26,6 +29,34 @@ type Engine struct {
...
+	// QueryErrorCounter describes a metric that accumulates number of failed queries monotonically.
+	QueryErrorCounter = discard.NewCounter()
+
+	// QueryHistogram describes a queries latency.
+	QueryHistogram = discard.NewHistogram()
+)
+
+func observeQuery(ctx *sql.Context, query string) func(err error) {
+	logrus.WithField("query", query).Debug("executing query")
+	span, _ := ctx.Span("query", opentracing.Tag{Key: "query", Value: query})
```
juanjux: It's safe to ignore the possible error here?

kuba--: it's not an error, the second returned value is context

_example/main.go
```diff
@@ -64,3 +81,86 @@ func createTestDatabase() *mem.Database {
 	table.Insert(ctx, sql.NewRow("Evil Bob", "evilbob@gmail.com", []string{"555-666-555", "666-666-666"}, time.Now()))
 	return db
 }
+
+func initMetrics() {
```
erizocosmico: Do we want to add all this extra code in the example? It seems like a lot of code not related to the example to have there. Put it in the README to let the user know how to customize the metrics? Yes. But here does not make much sense imho.

ajnavarro:
@kuba-- could you rebase please?

PR #745, opened 2019-06-05
Replace fork go-vitess by original vitess.

kuba--:
This PR replaces fork `gopkg.in/src-d/go-vitess.v1` by original vitess `vitess.io/vitess/go`.
The new `log.go` file in `sqle` package just overwrites vitess `log` implementation (replaces glog by logrus). We can keep the file till vitess implements logging through some easy to custom interface.


kuba--:
I suggest to archive `go-vitess` repo.

PR #750, opened 2019-06-12
sql/analyzer: refactor resolve_natural_joins rule

erizocosmico:
Now it's more simple, easier to understand and has way less code.


juanjux:
Good shaving you did there.

PR #753, opened 2019-06-18
sql: correctly handle nulls in SQL type conversion

erizocosmico:
Will fix https://github.com/src-d/gitbase/issues/878

Apparently, we've been handling nulls the wrong way all along when converting data to send to MySQL.


ajnavarro:
@erizocosmico could you have a look to Travis? I'm facing the same error with:
`SELECT uast_extract(NULL, "Go") from dual;`

erizocosmico:
Should be fixed now. One test was not correctly adapted.

Issue #755, opened 2019-06-20
Panic on query using EXPLODE


juanjux:
Adding it here because I think it's related to explode.

The query:

```sql
SELECT
    EXPLODE(UAST_EXTRACT(UAST(f.blob_content, 'Go', '//uast:Import/Path'), 'Value')) 
FROM files as f
WHERE f.file_path REGEXP('.*.go$')
    AND f.file_path NOT REGEXP '^vendor.*'
    AND NOT IS_BINARY(f.blob_content)
    AND f.blob_size < 1000000
LIMIT 1;
```

Causes this panic:

```
ERRO[0703] mysql_server caught panic:
unresolved column is a placeholder node, but IsNullable was called
/home/juanjux/go/pkg/mod/github.com/src-d/go-mysql-server@v0.4.1-0.20190619104848-eaab1795353a/sql/expression/unresolved.go:41 (0x9e7148)
	com/src-d/go-mysql-server/sql/expression.(*UnresolvedColumn).IsNullable: panic("unresolved column is a placeholder node, but IsNullable was called")
/home/juanjux/sourced/gitbase/internal/function/uast.go:68 (0xea4bb7)
	com/src-d/gitbase/internal/function.(*uastFunc).IsNullable: return u.Blob.IsNullable() || u.Mode.IsNullable() ||
/home/juanjux/go/pkg/mod/github.com/src-d/go-mysql-server@v0.4.1-0.20190619104848-eaab1795353a/sql/expression/common.go:55 (0x9dbc82)
	com/src-d/go-mysql-server/sql/expression.(*BinaryExpression).IsNullable: return p.Left.IsNullable() || p.Right.IsNullable()
/home/juanjux/go/pkg/mod/github.com/src-d/go-mysql-server@v0.4.1-0.20190619104848-eaab1795353a/sql/expression/function/explode.go:71 (0xcdd2a2)
	com/src-d/go-mysql-server/sql/expression/function.(*Generate).IsNullable: func (e *Generate) IsNullable() bool { return e.Child.IsNullable() }
/home/juanjux/go/pkg/mod/github.com/src-d/go-mysql-server@v0.4.1-0.20190619104848-eaab1795353a/sql/analyzer/resolve_generators.go:47 (0xd1e410)
	com/src-d/go-mysql-server/sql/analyzer.resolveGenerators.func1: expression.NewGetField(g.idx, g.expr.Type(), name, g.expr.IsNullable()),
[...]

PR #765, opened 2019-06-24
function: correctly transform up explode nodes

erizocosmico:
Will fix https://github.com/src-d/go-mysql-server/issues/755

Same as the PR a few days ago, but in the placeholder node. What I don't know is why tests passed without this but the query on Gitbase didn't 🤷‍♂


ajnavarro:
Can we add an integration test using explode + split + some column to try to reproduce the error case?

PR #768, opened 2019-06-24
sql: use custom boolean coercion logic in condition evaluation

erizocosmico:
Part of the fix for https://github.com/src-d/gitbase/issues/902

The idea is to use `EvaluateCondition` as a common logic for filtering in both gitbase squashed tables and go-mysql-server.

Right now, in go-mysql-server we just check the result `== true`, which not only is not correct but is not consistent with how gitbase does it.
In gitbase we use `sql.Boolean.Convert`, but by definition the convert method of boolean can't handle some types.

That's why I added some helper function for this with the type coercion logic that we can also reuse on gitbase.

### DIFF
--- a/sql/core.go
+++ b/sql/core.go
@@ -3,6 +3,8 @@ package sql // import "github.com/src-d/go-mysql-server/sql"
 import (
 	"fmt"
 	"io"
+	"math"
+	"time"
 
 	"gopkg.in/src-d/go-errors.v1"
 )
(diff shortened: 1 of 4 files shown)


sql/core.go
```diff
@@ -218,3 +220,27 @@ type Lockable interface {
...
+	if err != nil {
+		return false, err
+	}
+
+	switch b := v.(type) {
+	case bool:
+		return b, nil
+	case int, int64, int32, int16, int8, uint, uint64, uint32, uint16, uint8:
+		return b != 0, nil
+	case time.Duration:
```
juanjux: Should a Duration or time really ever evaluate to false?

erizocosmico: when it's 0, i guess

sql/core.go
```diff
@@ -218,3 +220,27 @@ type Lockable interface {
...
+		return int64(b) != 0, nil
+	case time.Time:
+		return b.UnixNano() != 0, nil
+	case float32, float64:
+		return int(math.Round(v.(float64))) != 0, nil
+	default:
+		return false, nil
```
kuba--: how about the string:
`select 1 from dual where  "1";`
basically it false, but for above string it's true for mysql.

sql/core.go
juanjux: But those are valid dates/durations. Anyway it's very minor; we should probably just do whatever MySQL do in these cases.

PR #773, opened 2019-06-26
sql: implement new API for node transformation

### LINKED ISSUE #772: Reduce amount of implemented code on Expressions
To reduce the amount of implemented code on Expressions and Nodes, we can heavily use UnaryExpression, BinaryExpression, and so on. 

Right now, these interfaces are just implementing `Children()`, `Resolved()` and `IsNullable()`.

We could implement also `TransformUp` function as following:
- Create a new interface in charge of return a new instance of the implemented expression:
```
type ExpressionInstanceGenerator interface {
	Generate(exprs ...sql.Expression) (sql.Expression, error)
}
```

- With this, we will be able to implement this interface on Unary and Binary expressions:
```
type BinaryExpression struct {
	ExpressionInstanceGenerator
	Left  sql.Expression
	Right sql.Expression
}
```

- After that, `TransformUp` method can be implemented on BinaryExpression. `Generate` method is the one that will be implemented on the other Expressions:
```
func (p *BinaryExpression) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
	var err error

	var l sql.Expression
	if p.Left != nil {
		l, err = p.Left.TransformUp(f)
		if err != nil {
			return nil, err
		}
	}
	var r sql.Expression
	if p.Left != nil {
		r, err = p.Right.TransformUp(f)
		if err != nil {
			return nil, err
		}
	}

	exp, err := p.Generate(l, r)
	if err != nil {
		return nil, err
	}

	return f(exp)
}
```

Doing this, we will reduce bugs on bad implemented TransformUp functions, also we can create generic Tests suites for this kind of expressions.

erizocosmico:
Close #772 

Instead of having TransformUp and TransformExpressionsUp in each
node, which was really error prone, now we have a different API.

Each node will have a WithChildren method that will receive
the children from which a new node of the same type will be created.
These nodes must come in the same number and order as the ones
returned by the Children method.

Expressioner nodes will also have WithExpressions method, which is
the same, except it will create a new node with its expressions
changed, instead of the children nodes.

The plan package will expose 3 new helpers:

- TransformUp: which transforms a node from the bottom-up.
- TransformExpressionsUp: which transforms expressions of a node from
  the bottom up.
- TransformExpressions: which transforms the expressions of just the
  given node.

Just like with nodes, expressions will also have a new WithChildren
method that does the exact same thing as it does in the nodes.

The expression package will expose a new helper:

- TransformUp: which transforms an expression from the bottom up.

The sql package now has one interface more: `OpaqueNode`. An opaque node is a node that acts as a black box and their children will not be transformed using the aforementioned helpers. SubqueryAlias node, for example, requires being opaque.

Caveats and limitations:

One thing that may seem odd is the limitation that WithChildren and
WithExpressions must receive the children in the exact same order and
number as they were returned from Expressions or Children.

This is because without this limitation there is no way to build
certain nodes. If we force this limitation on one, it would feel odd
not to have it elsewhere.

For example, take Case expression into account. It may or may not have
Expr, it has a list of branches (each having 2 expressions) and it
may or may not have an Else expression. If WithChildren receives N
children, how do we know where to put all these expressions?
This limitation allows us to build the node beacause we know the shape
of children must match the current shape.

### DIFF
--- a/sql/core.go
+++ b/sql/core.go
@@ -25,6 +25,10 @@ var (
 
 	//ErrUnexpectedRowLength is thrown when the obtained row has more columns than the schema
 	ErrUnexpectedRowLength = errors.NewKind("expected %d values, got %d")
+
+	// ErrInvalidChildrenNumber is returned when the WithChildren method of a
+	// node or expression is called with an invalid number of arguments.
+	ErrInvalidChildrenNumber = errors.NewKind("%T: invalid children number, got %d, expected %d")
 )
 
 // Nameable is something that has a name.
@@ -45,17 +49,6 @@ type Resolvable interface {
 	Resolved() bool
 }
 
-// Transformable is a node which can be transformed.
-type Transformable interface {
-	// TransformUp transforms all nodes and returns the result of this transformation.
-	// Transformation is not propagated to subqueries.
-	TransformUp(TransformNodeFunc) (Node, error)
-	// TransformExpressionsUp transforms all expressions inside the node and all its
-	// children and returns a node with the result of the transformations.
-	// Transformation is not propagated to subqueries.
-	TransformExpressionsUp(TransformExprFunc) (Node, error)
-}
-
 // TransformNodeFunc is a function that given a node will return that node
 // as is or transformed along with an error, if any.
 type TransformNodeFunc func(Node) (Node, error)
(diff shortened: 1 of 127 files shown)


sql/core.go
```diff
@@ -100,24 +95,37 @@ type Aggregation interface {
...
 	Resolvable
-	Transformable
 	fmt.Stringer
 	// Schema of the node.
 	Schema() Schema
 	// Children nodes.
 	Children() []Node
 	// RowIter produces a row iterator from this node.
 	RowIter(*Context) (RowIter, error)
+	// WithChildren returns a copy of the node with children replaced.
```
juanjux: Maybe "transformed" would be better here instead of replaced?

erizocosmico: Children may or may have not been transformed. This creates a new node with the new children in the places where previous children used to be.

In Opaque the docs refer to "transform" because a node that is opaque will not have its children transformed in TransformUp, TransformExpressions, etc. WithChildren and Children do not have anything to do with that.

On one hand we have the transform concept, which uses Children to get the children, then transforms them and replaces the old with the new ones using WithChildren.
On the other hand, we have replacement, which is what WithChildren is all about.

ajnavarro:
@erizocosmico could you rebase please?

erizocosmico:
Done

erizocosmico:
Updated so that resolve_having can take advantage of the new `WithChildren` API

PR #781, opened 2019-07-03
server: start root trace span on context creation

### LINKED ISSUE #778: No root span created for a query
Hi,

When I made a sql query (like `select * from sql where id = 1`), I got four traces (see [screenshot](http://ww4.sinaimg.cn/large/006tNc79ly1g4krjj1p9cj313w0jjgnv.jpg)). I expected that there was a single trace that represented the whole execution path for this query, not four separate traces each of which represented a phase of the execution path.

According to the source code [https://github.com/src-d/go-mysql-server/blob/master/server/handler.go#L78-L90](https://github.com/src-d/go-mysql-server/blob/master/server/handler.go#L78-L90),  a context `ctx` is created and is passed into `Query`. There is no root span created in between. 

Is there a way to create a root span without changing the source code? Thanks!

erizocosmico:
Fixes #778

### DIFF
--- a/server/context.go
+++ b/server/context.go
@@ -89,7 +89,10 @@ func (s *SessionManager) NewContextWithQuery(
 	s.mu.Unlock()
 
 	context := sql.NewContext(
-		context.Background(),
+		opentracing.ContextWithSpan(
+			context.Background(),
+			s.tracer.StartSpan("query"),
+		),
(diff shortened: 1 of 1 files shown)


ajnavarro:
Ignoring error on openjdk10 job.

PR #783, opened 2019-07-03
sql: fix how nulls are sent to clients

### LINKED ISSUE #779: NULL is not printed in the resulting table
Hi,

When I used mysql client to connect the SQL server, I got the following table in which there were some empty fields that were supposed to be printed as `NULL`.

What I got: 
```bash
$ mysql -h 127.0.0.1 -u root
mysql> select * from tbl where timestamp = 3;
+--------------+-----------+-----------+-------------+
| id           | timestamp | col1      | col2       |
+--------------+-----------+-----------+-------------+
| id1          |         3 |           |           3 |
| id2          |         3 |           |           3 |
| id3          |         3 |         3 |             |
+--------------+-----------+-----------+-------------+
3 rows in set (0.51 sec)
```

What I expect: 
```bash
$ mysql -h 127.0.0.1 -u root
mysql> select * from tbl where timestamp = 3;
+--------------+-----------+-----------+-------------+
| id           | timestamp | col1      | col2        |
+--------------+-----------+-----------+-------------+
| id1          |         3 |   NULL    |           3 |
| id2          |         3 |    NULL   |           3 |
| id3          |         3 |         3 |    NULL     |
+--------------+-----------+-----------+-------------+
3 rows in set (0.51 sec)
```

It seems to me that in https://github.com/src-d/go-mysql-server/pull/753/files#diff-654bac6a55597b15c4a5d404095b4e69R306, shouldn't we handle both `nil` and `nullT`? For example:

```go
if _, ok := v.(nullT); ok || v == nil {
  return sqltypes.NULL, nil
}
```

erizocosmico:
Fixes #779 

Fixing this, connecting from MySQL client broke. After much debugging, turns out it was because `@@version` and `@@version_comment` didn't exist (and thus, they were null), but the client was expecting something there. I just added both, but left them empty and the problem goes away.

### DIFF
--- a/sql/session.go
+++ b/sql/session.go
@@ -175,6 +175,8 @@ func DefaultSessionConfig() map[string]TypedValue {
 		"ndbinfo_version":          TypedValue{Text, ""},
 		"sql_select_limit":         TypedValue{Int32, math.MaxInt32},
 		"transaction_isolation":    TypedValue{Text, "READ UNCOMMITTED"},
+		"version":                  TypedValue{Text, ""},
+		"version_comment":          TypedValue{Text, ""},
 	}
 }
 
--- a/_integration/go/mysql_test.go
+++ b/_integration/go/mysql_test.go
@@ -74,7 +74,7 @@ func TestGrafana(t *testing.T) {
 		},
 		{
 			`select @@version_comment limit 1`,
-			[][]string{{"NULL"}},
+			[][]string{{""}},
 		},
 		{
 			`describe table mytable`,
--- a/engine_test.go
+++ b/engine_test.go
@@ -634,6 +634,8 @@ var queries = []struct {
 			{"ndbinfo_version", ""},
 			{"sql_select_limit", math.MaxInt32},
 			{"transaction_isolation", "READ UNCOMMITTED"},
+			{"version", ""},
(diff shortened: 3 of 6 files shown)


sql/session.go
```diff
@@ -175,6 +175,8 @@ func DefaultSessionConfig() map[string]TypedValue {
 		"ndbinfo_version":          TypedValue{Text, ""},
 		"sql_select_limit":         TypedValue{Int32, math.MaxInt32},
 		"transaction_isolation":    TypedValue{Text, "READ UNCOMMITTED"},
+		"version":                  TypedValue{Text, ""},
```
ajnavarro: Maybe we can add the real version here to maintain consistency.

erizocosmico: What version should we put? A fake mysql version? It might not be consistent with `VERSION()` though

ajnavarro: we cannot use the same param as VERSION() is using here?

erizocosmico: That's the thing, VERSION() doesn't even exist in plain go-mysql-server. It's registered in gitbase with a specific version.

ajnavarro: oh, right, I mix both. Sorry for the noise.

PR #784, opened 2019-07-03
Zachmu/bug fixes to contribute

zachmu:
Several bug fixes and improvements:

1) Float comparisons were broken because they were being truncated to integers. Fixed and added many tests with float values.
2) Negative float values were getting optimized away to 0 by the parser, fixed and added tests.
3) Added integration tests of IS NULL, IS NOT NULL
4) Implemented IS TRUE, IS FALSE, and NOT versions of each
5) Fixed some typos in the source
6) Allow conversion of string > bool to match MySQL semantics
7) Fixed bug where some non-column filter expressions were getting effectively optimized away by being force-converted to a boolean

### DIFF
--- a/sql/analyzer/optimization_rules.go
+++ b/sql/analyzer/optimization_rules.go
@@ -390,6 +390,16 @@ func evalFilter(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) 
 					return e, nil
 				}
 
+				// UnaryMinus expressions come back from the parser when a negative float is evaluated. Treat them just like
+				// normal literal expressions.
+				if um, ok := e.(*expression.UnaryMinus); ok {
+					negated, err := e.Eval(ctx, nil)
+					if err != nil {
+						return nil, err
+					}
+					return expression.NewLiteral(negated, um.Type()), nil
+				}
+
 				val, err := e.Eval(ctx, nil)
 				if err != nil {
 					return e, nil
--- a/sql/analyzer/optimization_rules.go
+++ b/sql/analyzer/optimization_rules.go
@@ -381,36 +381,19 @@ func evalFilter(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error) 
 				}
 
 				return e, nil
+			case *expression.Literal, expression.Tuple:
+				return e, nil
 			default:
 				if !isEvaluable(e) {
 					return e, nil
 				}
 
-				if _, ok := e.(*expression.Literal); ok {
-					return e, nil
-				}
-
-				// UnaryMinus expressions come back from the parser when a negative float is evaluated. Treat them just like
-				// normal literal expressions.
-				if um, ok := e.(*expression.UnaryMinus); ok {
-					negated, err := e.Eval(ctx, nil)
-					if err != nil {
-						return nil, err
-					}
-					return expression.NewLiteral(negated, um.Type()), nil
-				}
-
+				// All other expressions types can be evaluated once and turned into literals for the rest of query execution
 				val, err := e.Eval(ctx, nil)
 				if err != nil {
 					return e, nil
 				}
-
-				val, err = sql.Boolean.Convert(val)
-				if err != nil {
-					return e, nil
-				}
-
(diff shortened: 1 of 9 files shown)


zachmu:
Sorry for the relatively messy diff (includes several recent changes from your master), not sure what caused it exactly. Open to suggestions for how to make it cleaner.

zachmu:
Cleaned this up with the help of my colleague, ready to be reviewed.

sql/expression/istrue.go
```diff
@@ -0,0 +1,77 @@
...
+
+// TransformUp implements the Expression interface.
+func (e *IsTrue) TransformUp(f sql.TransformExprFunc) (sql.Expression, error) {
+	child, err := e.Child.TransformUp(f)
+	if err != nil {
+		return nil, err
+	}
+	if e.invert {
+		return f(NewIsFalse(child))
+	} else {
```
erizocosmico: You can just remove the else branch entirely, since the if branch returns

sql/expression/istrue.go
```diff
@@ -0,0 +1,77 @@
+package expression
+
+import "github.com/src-d/go-mysql-server/sql"
+
+// IsNull is an expression that checks if an expression is true.
```
juanjux: s/IsNull/IsTrue

sql/expression/istrue_test.go
```diff
@@ -0,0 +1,49 @@
...
+	require.Equal(false, eval(t, e, sql.NewRow(false)))
+
+	intF := NewGetField(0, sql.Int64, "col1", true)
+	e = NewIsTrue(intF)
+	require.Equal(sql.Boolean, e.Type())
+	require.False(e.IsNullable())
+	require.Equal(false, eval(t, e, sql.NewRow(nil)))
+	require.Equal(true, eval(t, e, sql.NewRow(100)))
+	require.Equal(true, eval(t, e, sql.NewRow(-1)))
+	require.Equal(false, eval(t, e, sql.NewRow(0)))
```
juanjux: Could you add some tests with other data types (like strings)?

juanjux:
Only a couple easy to fix things, thanks for the PR!

sql/analyzer/optimization_rules.go
```diff
@@ -389,6 +389,16 @@ func evalFilter(ctx *sql.Context, a *Analyzer, node sql.Node) (sql.Node, error)
 					return e, nil
 				}
 
+				// UnaryMinus expressions come back from the parser when a negative float is evaluated. Treat them just like
+				// normal literal expressions.
+				if um, ok := e.(*expression.UnaryMinus); ok {
+					negated, err := e.Eval(ctx, nil)
```
ajnavarro: Can we have here the case that UnaryMinus is wrapping a non literal child?

sql/expression/istrue.go
zachmu: Done.

sql/expression/istrue.go
zachmu: Done.

sql/analyzer/optimization_rules.go
zachmu: That should be covered by the isEvaluable check above (returns false if an expression contains columns). I don't want to assert a Literal child here because it wouldn't work for weird arithmetic expressions e.g. -(60*60*24).

sql/expression/istrue_test.go
zachmu: Done. Also caught a bug related to nil values :D

zachmu:
Addressed PR feedback and updated to include the recent interface change to Expression.

Adding tests for boolean columns for IsTrue evaluation as Juanjo suggested required being able to evaluate strings as booleans. Once implemented, this exposed a bug in the filter optimizer, where non-column expressions were incorrectly being coerced to booleans. Really, this is the same bug that was causing UnaryMinus expressions (for negative floats) to not work correctly. My fix is more general than my original PR: don't coerce to boolean during optimization. This is slightly less performant but more correct.

Please take another look and merge if it looks good.

juanjux:
Thanks for the fixes!

PR #788, opened 2019-07-09
sql/expression: correctly evaluate conditions in CASE

### LINKED ISSUE #787: Error on CASE statements when result is not bool
This query:

```
select CASE WHEN NULL THEN "yes" ELSE "no" END AS test
```

Is throwing the following error:

`ERROR 1105 (HY000): unknown error: unable to cast nil to bool`

But it should return `no`

erizocosmico:
Fixes #787

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1295,6 +1295,10 @@ var queries = []struct {
 		`SELECT COUNT(DISTINCT t.i) FROM tabletest t, mytable t2`,
 		[]sql.Row{{int64(3)}},
 	},
+	{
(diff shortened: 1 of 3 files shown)


ajnavarro:
Shall we review other similar cases to try to catch more places where this bug might be happening?

erizocosmico:
Done. No more usages of `sql.Boolean.Convert` in the wild

erizocosmico:
Reverting, the sql.Boolean.Convert in an optimization rule was correct

PR #794, opened 2019-08-05
function: implement regexp_matches

### LINKED ISSUE #756: Support for regexp_matches
Currently when there isn't a parser available for a file type, it would be very useful to be able to do substring extractions based on regular expressions.

The use case I was trying was to build a table of base images use in Dockerfiles. In Postgres I would use the following function for this.

```
regexp_matches(string text,pattern text [, flags text]) | setof text[] | Return all captured substrings resulting from matching a POSIX regular expression against the string. See Section 9.7.3 for more information. | regexp_matches('foobarbequebaz', '(bar)(beque)') | {bar,beque}
```
source: https://www.postgresql.org/docs/9.1/functions-string.html

In gitbase I can imagine this working like this:

```
SELECT
    r.repository_id AS repo,
    c.committer_when AS date,
    file_path AS path,
    EXPLODE(REGEXP_MATCHES( f.blob_content,'FROM ([^\s]+)')) AS base_images
FROM
    refs AS r
    NATURAL JOIN commits AS c
    NATURAL JOIN commit_files AS cm
    NATURAL JOIN files AS f
WHERE r.ref_name = 'HEAD'
    AND f.file_path REGEXP('Dockerfile')
    AND NOT IS_BINARY(f.blob_content)
    AND f.blob_size < 1000000
    AND f.file_path NOT REGEXP 'vendor.*'
    AND f.blob_content REGEXP 'FROM.*'
```

erizocosmico:
Closes #756

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1454,6 +1454,34 @@ var queries = []struct {
 		ORDER BY table_type, table_schema, table_name`,
 		[]sql.Row{{"mydb", "mytable", "TABLE"}},
 	},
+	{
+		`SELECT REGEXP_MATCHES("bopbeepbop", "bop")`,
+		[]sql.Row{{[]interface{}{"bop", "bop"}}},
+	},
+	{
+		`SELECT EXPLODE(REGEXP_MATCHES("bopbeepbop", "bop"))`,
+		[]sql.Row{{"bop"}, {"bop"}},
+	},
+	{
+		`SELECT EXPLODE(REGEXP_MATCHES("helloworld", "bop"))`,
+		[]sql.Row{},
+	},
+	{
+		`SELECT EXPLODE(REGEXP_MATCHES("", ""))`,
+		[]sql.Row{{""}},
+	},
+	{
+		`SELECT REGEXP_MATCHES(NULL, "")`,
+		[]sql.Row{{nil}},
+	},
+	{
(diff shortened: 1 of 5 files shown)


engine_test.go
```diff
@@ -1454,6 +1454,18 @@ var queries = []struct {
...
+		[]sql.Row{{[]interface{}{"bop", "bop"}}},
+	},
+	{
+		`SELECT EXPLODE(REGEXP_MATCHES("bopbeepbop", "bop"))`,
+		[]sql.Row{{"bop"}, {"bop"}},
+	},
+	{
+		`SELECT EXPLODE(REGEXP_MATCHES("helloworld", "bop"))`,
+		[]sql.Row{},
+	},
```
juanjux: I would add empty + nil tests just in case.

sql/expression/function/regexp_matches_test.go
```diff
@@ -0,0 +1,111 @@
...
+			nil,
+		},
+		{
+			"bop",
+			"bopbeepBop",
+			"ix",
+			nil,
+			errInvalidRegexpFlag,
+		},
+	}
```
juanjux: ditto

erizocosmico:
Done @juanjux

erizocosmico:
@ajnavarro it was a temporary error downloading dependencies, restarted the failing job

PR #798, opened 2019-08-07
Make Sleep check for cancelled context every second

### LINKED ISSUE #797: sleep function cannot be cancelled
we are just calling to time.Sleep, and that cannot be canceled from the Context. Change sleep to make it cancellable.

juanjux:
Fixes #797

### DIFF
--- a/sql/expression/function/sleep.go
+++ b/sql/expression/function/sleep.go
@@ -37,29 +37,12 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		return nil, err
 	}
 
-	fchild := child.(float64)
-	if fchild <= 0 {
+	select {
+	case <-time.After(time.Duration(child.(float64) * 1000) * time.Millisecond):
+		return 0, nil
+	case <-ctx.Done():
 		return 0, nil
 	}
-
-	// Wake up every second to check if the context was cancelled
-	remaining := fchild * 1000.0
-	for remaining >= 0 {
-		toSleep := 1000.0
-		if remaining < 1000 {
-			toSleep = remaining
-		}
-		remaining -= 1000
-
-		select {
-		case <-ctx.Done():
-			goto End
-		case <-time.After(time.Duration(toSleep) * time.Millisecond):
-		}
-	}
-End:
-
-	return 0, nil
 }
 
 // String implements the Stringer interface.
--- a/sql/expression/function/sleep.go
+++ b/sql/expression/function/sleep.go
@@ -37,7 +37,28 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		return nil, err
 	}
 
-	time.Sleep(time.Duration(child.(float64)*1000) * time.Millisecond)
+	fchild := child.(float64)
+	if fchild <= 0 {
+		return 0, nil
+	}
+
+	// Wake up every second to check if the context was cancelled
+	remaining := fchild * 1000.0
+	for remaining >= 0 {
+		toSleep := 1000.0
+		if remaining < 1000 {
+			toSleep = remaining
+		}
+		remaining -= 1000
+
+		select {
+		case <-ctx.Done():
+			goto End
+		case <-time.After(time.Duration(toSleep) * time.Millisecond):
+		}
+	}
+End:
+
 	return 0, nil
 }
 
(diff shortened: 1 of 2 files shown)


juanjux:
Updated to use better solution by @creachadair.

sql/expression/function/sleep.go
```diff
@@ -37,8 +37,12 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		return nil, err
 	}
 
-	time.Sleep(time.Duration(child.(float64)*1000) * time.Millisecond)
-	return 0, nil
+	select {
+	case <-time.After(time.Duration(child.(float64)) * 1000.0):
```
creachadair: So I think this implies that `child` is in microseconds? To make that more self-documenting, I suggest:

```go
time.Duration(child.(float64))*time.Microsecond
```

juanjux: `child` is in seconds but mysql takes decimals so we convert to milliseconds.

creachadair: Right, ok, but `time.After` and `time.Sleep` take nanoseconds (which is what `time.Duration` is), so multiplying by 1000 is definitely not right.

juanjux: Are you sure about that? It works...

juanjux: Oh frag, I notices it doesn't have the `time.Milliseconds` in the version in this PR (I was testing with gitbase so modifiying in another branch, that's why it worked, my version had it).

creachadair: Ah, if it's seconds then you want `time.Duration(child.(float64))*time.Second`, that will give you the right scaling without an extra factor.

creachadair: (Or if you're concerned about fractions, then you probably want the multiplier _before_ the conversion to int: `time.Duration(child.(float64) * 1000) * time.Millisecond` for example).

juanjux: Multiplying by time.seconds seems to floor() it:

```

Comment (juanjux) [reply sql/expression/function/sleep.go]: Added the 1000 inside the parents, thanks again!

Comment (creachadair) [reply sql/expression/function/sleep.go]: You'll still want the `time.Milliseconds` scaling factor outside the parens, though, after the conversion to `time.Duration`.

File: sql/expression/function/sleep.go [hunk 2]
```diff
@@ -37,8 +37,12 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		return nil, err
 	}
 
-	time.Sleep(time.Duration(child.(float64)*1000) * time.Millisecond)
-	return 0, nil
+	select {
+	case <-time.After(time.Duration(child.(float64) * 1000) * time.Millisecond):
```
erizocosmico: I would create a timer using `time.NewTimer` instead and then `defer timer.Stop()` so that we don't leave potentially long-running timers running unnecessarily.

sql/expression/function/sleep.go
```diff
@@ -37,8 +37,12 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
...
+		return 0, nil
+	case <-ctx.Done():
+		return 0, nil
```
erizocosmico: I would return `context.Canceled` here as the error

sql/expression/function/sleep.go
juanjux: Done.

sql/expression/function/sleep.go
juanjux: Done.

sql/expression/function/sleep.go
```diff
@@ -37,8 +37,15 @@ func (s *Sleep) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
...
 	}
 
-	time.Sleep(time.Duration(child.(float64)*1000) * time.Millisecond)
-	return 0, nil
+	t := time.NewTimer(time.Duration(child.(float64) * 1000) * time.Millisecond)
+	defer t.Stop()
+
+	select {
+	case <-ctx.Done():
+		return context.Canceled, nil
```
erizocosmico: context.Canceled should be returned as the last parameter, which is the error

juanjux: Thanks, fixed.

Issue #800, opened 2019-08-09
Long processing queries don't stop if the connection is dead/killed until the timeout triggers


juanjux:
Related to: 
https://github.com/src-d/gitbase-spark-connector-enterprise/issues/81

If a connection is dead (for example because the client was killed) queries that take a long time to send anything back will continue happily running until the configured or OS timeout triggers, which could be a long time. We should be able to detect a dead connection and stop the running query.

This is a hairy problem because AFAIK we can't detect if a socket is dead without doing a test reading or writing from it, and that could interfere with the real readings/writings, specially if done on a goroutine as it probably should. Maybe reading and seeking one position back could work (but also could explode spectacularly).

The best option, if it worked, would be using a `mysql.Conn.Ping()` but if run from a goroutine it gives and error about reset counters (because the implementation [really resets them](https://github.com/vitessio/vitess/blob/master/go/mysql/client.go#L174) so it looks like we can't call it in parallel with a running query.

PR #801, opened 2019-08-09
Give vitess a chance to enforce connection timeouts if we've been waiting for a row for longer

### LINKED ISSUE #799: Connection timeouts can't be processed if a query is running.
Connection timeouts from Vitess doesn't trigger, and thus stop the query, is a query is running but not sending anything. This is because query processing and the `Handler` run on the same thread, so when Vitess detects a timeout the call to `CloseConnection` can't be processed until the query finishes.

Related to: https://github.com/src-d/gitbase-spark-connector-enterprise/issues/81

juanjux:
Fixes #799 
Related issue: https://github.com/src-d/gitbase-spark-connector-enterprise/issues/81

This PR abort further `rowIterator.Next()` calls if the time waiting for a single row is bigger than the configured `ConnReadTimeout` thus giving a change to Vitess to call the `Handler.CloseConnection()` callback when it detects the timeout. It does it by running the calls to the iterator on a looped goroutine with an `Timer` which checks that we didn't exceed the timeout.

This doesn't fix issue #800 which implies detecting a closed connection before the timeout expires and killing the associated queries, but serves to somewhat workaround it by configuring a timeout.


server/handler.go
```diff
@@ -103,6 +106,30 @@ func (h *Handler) ComQuery(
...
+				return
+			}
+			rowc <- row
+		}
+	}(rowchan, errchan)
+
+	// Default (big) waitTime is 3600, but it wont matter if there is no timeout
+	// because the loop will just iterate again. If there is a timeout, it will
+	// be enforced to ensure that Vitess has a chance to call Handler.CloseConnection()
+	waitTime := 3600 * time.Second
```
creachadair: Could also be `waitTime := time.Hour`

server/handler.go
```diff
@@ -103,6 +106,30 @@ func (h *Handler) ComQuery(
...
+	go func(rowc chan sql.Row, errc chan error) {
+		for {
+			row, err := rows.Next()
+			if err != nil {
+				errc <- err
+				return
+			}
+			rowc <- row
+		}
+	}(rowchan, errchan)
```
creachadair: It isn't necessary to bind these to parameters—the values don't change during the lifetime of the closure.

```go
go func() {
   …
}()
```

server/handler.go
```diff
@@ -115,27 +142,34 @@ func (h *Handler) ComQuery(
...
 			continue
 		}
 
-		row, err := rows.Next()
-		if err != nil {
+		select {
+		case err := <-errchan:
 			if err == io.EOF {
-				break
+				goto EndFor
```
creachadair: Better is to label the `for` statement and break to that label:

```go
nextRow:
  for {
     …
     select {
        …
        break nextRow
```

server/handler.go
```diff
@@ -115,27 +142,34 @@ func (h *Handler) ComQuery(
...
+			if err != nil {
+				return err
+			}
 
-		outputRow, err := rowToSQL(schema, row)
-		if err != nil {
-			return err
+			r.Rows = append(r.Rows, outputRow)
+			r.RowsAffected++
+		case <-time.After(waitTime):
```
creachadair: Since the default is quite long, and the goroutine on the timer will not exit till the timer fires or is cancelled, you should probably consider cancelling this explicitly:

```go
timeout := time.NewTimer(waitTime)
defer timeout.Stop()
select {
…
case <-timeout.C:
```

server/handler.go
```diff
@@ -115,27 +142,34 @@ func (h *Handler) ComQuery(
...
-				break
+				goto EndFor
 			}
 
 			return err
-		}
+		case row := <-rowchan:
+			outputRow, err := rowToSQL(schema, row)
+			if err != nil {
+				return err
```
creachadair: Be careful: Returning at this point will orphan the goroutine that is reading from the rows channel, unless this is the last value in the iterator. The goroutine cannot exit because it will block trying to deliver a value and nothing will read any further values after this return. A goroutine cannot be GC'd until it exits. As a rule of thumb, try to make sure any goroutine you start will either run forever (the lifetime of the program) or has a well-defined way to return even in case of error.

Usually that means you either need to fully drain any channel you create, or ensure that the operation populating the channel is cancelled before exit. One way you might do that here:

```go
rctx, cancel := context.WithCancel(ctx)
defer cancel()
…
// in the goroutine
for {
	…
	select {
	case <-rctx.Done():
		return
	case rowc <- row:
		// ok
	}
```

This is imperfect, since it could still stall trying to write an error: You can work around that by buffering `errchan` with 1 slot.

server/handler.go
juanjux: True. You can get a programmer out of Python but you can't get the Python out of the programmer, it seems.

server/handler.go
juanjux: Done in next commit.

server/handler.go
juanjux: I like gotos for a little first world anarchy but that's cleaner.

server/handler.go
juanjux: I don't get you here, it doesn't matter if its long because it'll be cancelled as soon as there is another row or an error (or the timeout is exceeded). Or even if the `select` finish in one of the two other cases there will still be a hidden timer preventing the goroutine from being GC'ed?

server/handler.go
juanjux: This blocked for me, I used instead a signal channel to exit the goroutine in that case (with a buffer with 1 slot, because it could be busy getting the next row and block the sender), seems to work, I also refactored the logic a little to remote the `return`s from the `case`s which always make me feel uneasy, PTAL.

server/handler.go
juanjux: PS: To make my explanation clearer, I could use a much smaller waitTime (30 secs) and an accumulator for the total time to compare to the timeout to make it clearer that the wait in the `After` doesn't really matter (it will just loop over again until the timeout).

server/handler.go
creachadair: Hmm. It shouldn't block as long as at least one of those options can proceed. Under the covers, the context is really just using a signal channel anyway. What I had in mind was something along these lines: https://play.golang.org/p/LOM4jRDubmv

Anyway, though, I think a separate channel is also fine, but see my note on the other thread.

server/handler.go
creachadair: However, good point about the loop: Since it's a loop we can't safely defer (which doesn't happen till the function exits), but instead `timeout.Stop()` after the end of the select.

juanjux: Wow I didn't knew time.After would create a goroutine. We definitely don't what that, even short duration ones, because some queries could iterateoover thousands of rows pretty quickly, so I'll switch to a Timer.

creachadair: If you really only need one of them for the whole scan—rather than one timeout per row—you could move the timer creation outside the loop and re-use the channel inside the `select`, that way you get only 1 additional goroutine for the whole operation (and of course reusing the unexpired timer is fine, as you mentioned).

server/handler.go
```diff
@@ -109,27 +109,34 @@ func (h *Handler) ComQuery(
 
 	rowchan := make(chan sql.Row)
 	errchan := make(chan error)
+	quit := make(chan bool, 2)
```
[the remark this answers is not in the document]
juanjux: Done and works perfectly!

server/handler.go
juanjux: Optimization done! Let's reserve that RAM for Chrome :)

server/handler.go
```diff
@@ -103,6 +106,41 @@ func (h *Handler) ComQuery(
...
+			}
+			rowchan <- row
+		}
+	}()
+
+	// Default (big) waitTime is 3600, but it wont matter if there is no timeout
+	// because the loop will just iterate again. If there is a timeout, it will
+	// be enforced to ensure that Vitess has a chance to call Handler.CloseConnection()
+	totalTime := 0 * time.Second
+	waitTime := 10 * time.Second
```
erizocosmico: shouldn't this be 1 * time.Hour?

server/handler.go
```diff
@@ -103,6 +106,41 @@ func (h *Handler) ComQuery(
...
+				return
+			}
+			rowchan <- row
+		}
+	}()
+
+	// Default (big) waitTime is 3600, but it wont matter if there is no timeout
+	// because the loop will just iterate again. If there is a timeout, it will
+	// be enforced to ensure that Vitess has a chance to call Handler.CloseConnection()
+	totalTime := 0 * time.Second
```
erizocosmico: what's total time used for?

The only place it seems to be used is to assign waitTime to totalTime (which will make it be always equal to waitTime)

server/handler.go
juanjux: It was a test with smaller default timeouts that somewhat remained, changed and fixed.

server/handler.go
juanjux: Removed now, but it was an accumulator, on the loop of that commit there was a `totalTime += waitTime` it the timer expired and it would be checked to not exceed the connection timeout. Now if there is a timeout `waitTime` defaults to it or one hour if not (the time in this second case won't matter since it'll just loop again until the usually huge OS socket timeout expires).

server/handler.go
```diff
@@ -22,24 +22,27 @@ import (
 var regKillCmd = regexp.MustCompile(`^kill (?:(query|connection) )?(\d+)$`)
 
 var errConnectionNotFound = errors.NewKind("Connection not found: %c")
+var RowTimeout = errors.NewKind("Row read wait bigger than connection timeout")
```
erizocosmico: should be `ErrRowTimeout` and error text should start in lowercase

server/handler.go
```diff
@@ -115,26 +152,40 @@ func (h *Handler) ComQuery(
...
+				break rowLoop
+			}
 
-		outputRow, err := rowToSQL(schema, row)
-		if err != nil {
-			return err
+			r.Rows = append(r.Rows, outputRow)
+			r.RowsAffected++
+		case <-timer.C:
+			if h.readTimeout != 0 && waitTime >= h.readTimeout {
```
erizocosmico: what's the `waitTime >= h.readTimeout` part for?

server/handler.go
```diff
@@ -115,26 +152,40 @@ func (h *Handler) ComQuery(
...
+				// Return so Vitess can call the CloseConnection callback
+				err = RowTimeout.New()
+				close(quit)
+				break rowLoop
+			}
 		}
+		timer.Reset(waitTime)
+	}
+
+	timer.Stop()
```
erizocosmico: there are returns previous to this line. So, maybe it should be `defer timer.Stop()` just after its creation

server/handler.go
juanjux: Done. Also fixing the error before then since I guess it was wrong too not starting with uppercase too.

server/handler.go
juanjux: Vestigial, good catch, fixed.

server/handler.go
juanjux: Done.

server/handler.go
```diff
@@ -103,6 +106,41 @@ func (h *Handler) ComQuery(
...
+
+	// Default (big) waitTime is one hour if there is not timeout configured, in which case
+	// it will loop to iterate again. If there is a timeout, it will
+	// be enforced to ensure that Vitess has a chance to call Handler.CloseConnection()
+	waitTime := 1 * time.Hour
+
+	if h.readTimeout > 0 {
+		waitTime = h.readTimeout
+	}
+	timer := time.NewTimer(h.readTimeout)
```
erizocosmico: this should be waitTime instead of `h.readTimeout`, right?

juanjux: Right, I should avoid doing PRs on fridays, fixes.

juanjux: Right, done.

server/handler.go
```diff
@@ -115,26 +153,38 @@ func (h *Handler) ComQuery(
...
+				close(quit)
+				break rowLoop
+			}
 		}
+		timer.Reset(waitTime)
+	}
 
-		r.Rows = append(r.Rows, outputRow)
-		r.RowsAffected++
+	if err != nil {
```
erizocosmico: where does this err come from? I don't see any err defined in the scope

juanjux: It was for the third case in the select, but should have been removed when I added the `defer`, I've refactored it without it and direct returns on errors to make the select it less jumpy.

server/handler.go
```diff
@@ -21,25 +21,28 @@ import (
 
 var regKillCmd = regexp.MustCompile(`^kill (?:(query|connection) )?(\d+)$`)
 
-var errConnectionNotFound = errors.NewKind("Connection not found: %c")
+var errConnectionNotFound = errors.NewKind("connection not found: %c")
+var ErrRowTimeout = errors.NewKind("row read wait bigger than connection timeout")
```
creachadair: minor nit: This should have a doc comment, or be unexported.

server/handler.go
```diff
@@ -103,6 +106,41 @@ func (h *Handler) ComQuery(
 
 	var r *sqltypes.Result
 	var proccesedAtLeastOneBatch bool
+
+	rowchan := make(chan sql.Row)
+	errchan := make(chan error)
+	quit := make(chan struct{})
```
creachadair: [optional] You could put `defer close(quit)` here, instead of calling it in multiple places below. As far as I can see, anytime this function returns, you'd want `quit` closed anyway.

server/handler.go
```diff
@@ -115,26 +153,32 @@ func (h *Handler) ComQuery(
...
+				return err
+			}
 
-		outputRow, err := rowToSQL(schema, row)
-		if err != nil {
-			return err
+			r.Rows = append(r.Rows, outputRow)
+			r.RowsAffected++
+		case <-timer.C:
+			if h.readTimeout != 0 {
```
creachadair: Doesn't _any_ timeout count as an error? Assuming we went on long enough to overrun the default 1h timer shouldn't that also be an error?

server/handler_test.go
```diff
@@ -37,22 +38,23 @@ func setupMemDB(require *require.Assertions) *sqle.Engine {
 	return e
 }
 
+// This session builder is used as dummy mysql Conn is not complete and
+// causes panic when accessing remote address.
+func testSessionBuilder(c *mysql.Conn, addr string) sql.Session {
+	client := "127.0.0.1:34567"
```
creachadair:
```suggestion
	const client = "127.0.0.1:34567"
```

server/handler.go
juanjux: Done.

server/handler.go
juanjux: I prefer to make it explicit so the code reader can see where we are actively signaling the goroutine.

server/handler.go
juanjux: No, 0 means that there is no timeout configured for the connection, so the socket will (hopefully) timeout with the default OS one, which is usually huge (ideally this should detect cases where the socket is dead like when you kill the client, but it doesn't, see #800 . But on second thoughts the wait, and thus the check to not wait in vain, could be slower so I'm leaving it at 1 minute.

server/handler_test.go
juanjux: Done.

server/handler.go
```diff
@@ -103,6 +106,41 @@ func (h *Handler) ComQuery(
 
 	var r *sqltypes.Result
 	var proccesedAtLeastOneBatch bool
+
+	rowchan := make(chan sql.Row)
+	errchan := make(chan error)
+	quit := make(chan struct{})
+
+	// This goroutine will be select{}ed giving a change to Vitess to call the
```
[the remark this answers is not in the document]
juanjux: Fixed, thanks!

juanjux:
@ajnavarro dont merge, maybe my next PR will make this unnecesary.

juanjux:
Closing, definitely superseeded by the next PR that improves also this part and fixes a bug.

juanjux:
[This](https://github.com/src-d/go-mysql-server/pull/806) PR should fix the problem in Linux.

PR #802, opened 2019-08-12
sql: implement memory management system for caches

erizocosmico:
This PR implements a memory management system, intended to have
control over the allocated memory for caches so that they can be
freed at any moment and we can avoid out of memory errors.

The main changes are the following:

- MemoryManager in the sql package, which is just the component that
  tracks all caches. Memory of all freeable caches can be freed
  using the Free method of this component. The only way to instantiate
  new caches is using the NewXXXCache methods.
- Rows, history and LRU cache implementations, accessed using the
  NewXXXCache methods of MemoryManager.
- Reporters, which is a component that reports the maximum amount
  of memory the program is allowed to use and the currently used memory.
  This interface is meant for making testing easier. There is a default
  ProcessMemory reporter that returns the memory used by the process and
  the maximum memory defined in the `MAX_MEMORY` environment variable.
- MemoryManager is passed down to every component through *sql.Context,
  which meant a little more boilerplate on the server SessionBuilder.
- GroupBy, Sort, Distinct and Join now use the provided APIs of memory
  and cache management for their in-memory computations.

Caveats:
- We need to think of a good default so that memory usage won't grow
  forever and crash eventually, which is the behaviour when MAX_MEMORY is 0.

### DIFF
--- a/sql/cache.go
+++ b/sql/cache.go
@@ -1,0 +1,131 @@
+package sql
+
+import (
+	"fmt"
+	"hash/crc64"
+	"runtime"
+
+	lru "github.com/hashicorp/golang-lru"
+	errors "gopkg.in/src-d/go-errors.v1"
+)
+
+var table = crc64.MakeTable(crc64.ISO)
+
+// CacheKey returns a hash of the given value to be used as key in
+// a cache.
+func CacheKey(v interface{}) uint64 {
+	return crc64.Checksum([]byte(fmt.Sprintf("%#v", v)), table)
+}
+
+// ErrKeyNotFound is returned when the key could not be found in the cache.
+var ErrKeyNotFound = errors.NewKind("memory: key %d not found in cache")
+
+type lruCache struct {
+	memory   Freeable
+	reporter Reporter
+	size     int
+	cache    *lru.Cache
+}
+
+func newLRUCache(memory Freeable, r Reporter, size uint) *lruCache {
+	lru, _ := lru.New(int(size))
+	return &lruCache{memory, r, int(size), lru}
+}
+
+func (l *lruCache) Put(k uint64, v interface{}) error {
+	if releaseMemoryIfNeeded(l.reporter, l.Free, l.memory.Free) {
+		l.cache.Add(k, v)
+	}
+	return nil
+}
+
+func (l *lruCache) Get(k uint64) (interface{}, error) {
+	v, ok := l.cache.Get(k)
+	if !ok {
+		return nil, ErrKeyNotFound.New(k)
+	}
+
+	return v, nil
+}
+
+func (l *lruCache) Free() {
+	l.cache, _ = lru.New(l.size)
+}
+
+func (l *lruCache) Dispose() {
+	l.memory = nil
+	l.cache = nil
+}
+
+type rowsCache struct {
+	memory   Freeable
+	reporter Reporter
+	rows     []Row
+}
+
+func newRowsCache(memory Freeable, r Reporter) *rowsCache {
+	return &rowsCache{memory, r, nil}
+}
+
+func (c *rowsCache) Add(row Row) error {
+	if !releaseMemoryIfNeeded(c.reporter, c.memory.Free) {
+		return ErrNoMemoryAvailable.New()
+	}
+
+	c.rows = append(c.rows, row)
+	return nil
+}
+
+func (c *rowsCache) Get() []Row { return c.rows }
+
+func (c *rowsCache) Dispose() {
+	c.memory = nil
+	c.rows = nil
+}
+
+type historyCache struct {
+	memory   Freeable
+	reporter Reporter
+	cache    map[uint64]interface{}
+}
+
+func newHistoryCache(memory Freeable, r Reporter) *historyCache {
+	return &historyCache{memory, r, make(map[uint64]interface{})}
+}
+
+func (h *historyCache) Put(k uint64, v interface{}) error {
+	if !releaseMemoryIfNeeded(h.reporter, h.memory.Free) {
+		return ErrNoMemoryAvailable.New()
+	}
+	h.cache[k] = v
+	return nil
+}
+
+func (h *historyCache) Get(k uint64) (interface{}, error) {
+	v, ok := h.cache[k]
+	if !ok {
+		return nil, ErrKeyNotFound.New(k)
+	}
+	return v, nil
+}
+
+func (h *historyCache) Dispose() {
+	h.memory = nil
+	h.cache = nil
+}
+
+// releasesMemoryIfNeeded releases memory if needed using the following steps
+// until there is available memory. It returns whether or not there was
+// available memory after all the steps.
+func releaseMemoryIfNeeded(r Reporter, steps ...func()) bool {
+	for _, s := range steps {
+		if HasAvailableMemory(r) {
+			return true
+		}
+
+		s()
+		runtime.GC()
+	}
+
+	return HasAvailableMemory(r)
+}
--- a/sql/cache_test.go
+++ b/sql/cache_test.go
@@ -1,0 +1,169 @@
+package sql
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCacheKey(t *testing.T) {
+	k := CacheKey(1)
+	require.Equal(t, uint64(0x4320000000000000), k)
+}
+
+func TestLRUCache(t *testing.T) {
+	t.Run("basic methods", func(t *testing.T) {
+		require := require.New(t)
+
+		cache := newLRUCache(mockMemory{}, fixedReporter(5, 50), 10)
+
+		require.NoError(cache.Put(1, "foo"))
+		v, err := cache.Get(1)
+		require.NoError(err)
+		require.Equal("foo", v)
+
+		_, err = cache.Get(2)
+		require.Error(err)
+		require.True(ErrKeyNotFound.Is(err))
+
+		// Free the cache and check previous entry disappeared.
+		cache.Free()
+
+		_, err = cache.Get(1)
+		require.Error(err)
+		require.True(ErrKeyNotFound.Is(err))
+
+		cache.Dispose()
+		require.Panics(func() {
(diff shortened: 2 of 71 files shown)


sql/cache_test.go
```diff
@@ -0,0 +1,166 @@
...
+		// Free the cache and check previous entry disappeared.
+		cache.Free()
+
+		_, err = cache.Get(1)
+		require.Error(err)
+		require.True(ErrKeyNotFound.Is(err))
+
+		cache.Dispose()
+		require.Panics(func() {
+			cache.Get(1)
```
golangcibot: Error return value of `cache.Get` is not checked (from `errcheck`)

sql/cache_test.go
```diff
@@ -0,0 +1,166 @@
...
+		require.NoError(err)
+		require.Equal("foo", v)
+
+		_, err = cache.Get(2)
+		require.Error(err)
+		require.True(ErrKeyNotFound.Is(err))
+
+		cache.Dispose()
+		require.Panics(func() {
+			cache.Put(2, "foo")
```
golangcibot: Error return value of `cache.Put` is not checked (from `errcheck`)

sql/cache_test.go
```diff
@@ -0,0 +1,166 @@
...
+		require := require.New(t)
+
+		cache := newRowsCache(mockMemory{}, fixedReporter(5, 50))
+
+		require.NoError(cache.Add(Row{1}))
+		require.Len(cache.Get(), 1)
+
+		cache.Dispose()
+		require.Panics(func() {
+			cache.Add(Row{2})
```
golangcibot: Error return value of `cache.Add` is not checked (from `errcheck`)

sql/memory.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (m *MemoryManager) removeCache(pos uint64) {
+	m.mu.Lock()
+	delete(m.caches, pos)
+
+	if len(m.caches) == 0 {
+		m.token = 0
+	}
+
+	m.mu.Unlock()
```
agarciamontoro: Why isn't this `Unlock` deferred at the beginning as in `addCache`?

sql/cache.go
```diff
@@ -0,0 +1,131 @@
...
+	size   int
+	cache  *lru.Cache
+}
+
+func newLRUCache(memory Freeable, r Reporter, size int) *lruCache {
+	lru, _ := lru.New(size)
+	return &lruCache{memory, r, size, lru}
+}
+
+func (l *lruCache) Put(k uint64, v interface{}) error {
```
agarciamontoro: I saw  in other parts of the codebase that the functions implementing an interface had documentation as in "This implements KeyValue interface". I don't know if that's some kind of convention we have (is it?), but it definitely helps me when I am reading new code, so maybe it is worth it to write it here in `Put` and `Get`, as well as in `Free` and `Dispose` below.

sql/cache.go
```diff
@@ -0,0 +1,131 @@
...
+func CacheKey(v interface{}) uint64 {
+	return crc64.Checksum([]byte(fmt.Sprintf("%#v", v)), table)
+}
+
+// ErrKeyNotFound is returned when the key could not be found in the cache.
+var ErrKeyNotFound = errors.NewKind("memory: key %d not found in cache")
+
+type lruCache struct {
+	memory Freeable
+	r      Reporter
```
agarciamontoro: I'd expand the name of this variable a bit.

sql/cache.go
```diff
@@ -0,0 +1,131 @@
...
+		return ErrNoMemoryAvailable.New()
+	}
+	h.cache[k] = v
+	return nil
+}
+
+func (h *historyCache) Get(k uint64) (interface{}, error) {
+	v, ok := h.cache[k]
+	if !ok {
+		return nil, ErrKeyNotFound.New()
```
agarciamontoro: Should be `ErrKeyNotFound.New(k)`?

sql/cache.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (h *historyCache) Dispose() {
+	h.memory = nil
+	h.cache = nil
+}
+
+// releasesMemoryIfNeeded releases memory if needed using the following steps
+// until there is available memory. It returns whether or not there was
+// available memory after all the steps.
+func releaseMemoryIfNeeded(r Reporter, steps ...func()) bool {
```
agarciamontoro: This design is cool!

sql/cache_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+		require.NoError(cache.Put(1, "foo"))
+		_, err := cache.Get(1)
+		require.Error(err)
+		require.True(ErrKeyNotFound.Is(err))
+	})
+
+	t.Run("free required to add entry", func(t *testing.T) {
+		require := require.New(t)
+		var freed bool
```
agarciamontoro: Maybe we can explicitly assert that this variable is set to true after the Put? I understand that is implicit in the Get giving no error, but anyway. Same for the other tests below.

sql/cache.go
erizocosmico: yup, good catch

sql/cache_test.go
erizocosmico: Good catch. It was the intention of that variable but forgot to assert it later 😅

sql/cache.go
erizocosmico: Adding that doc to methods is basically so the linter doesn't complain, because it doesn't really add much value. The reason this is not commented is because the type is private, and thus the linter does not complain about it.

sql/memory.go
erizocosmico: Yep, probably better being deferred

sql/memory.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Reporter is a component that gives information about the memory usage.
+type Reporter interface {
+	// UsedMemory returns the memory in use in bytes.
+	MaxMemory() uint64
+	UsedMemory() uint64
+}
+
+// ProcessMemory is a reporter for the memory used by the process and the
+// maximum amount of money allowed controlled by the MAX_MEMORY environment
```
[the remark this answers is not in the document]
erizocosmico: oops

sql/memory.go
```diff
@@ -0,0 +1,189 @@
...
+const maxMemoryKey = "MAX_MEMORY"
+
+const (
+	b   = 1
+	kib = 1024 * b
+	mib = 1024 * kib
+)
+
+var maxMemory = func() uint64 {
+	v, _ := strconv.ParseUint(os.Getenv(maxMemoryKey), 10, 64)
```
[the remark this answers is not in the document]
erizocosmico: If it's not convertible, the function will return 0, which means not handled. On the other hand, it may actually be better to panic here if the value is misformatted, so the user does not think they set a max memory value and then be surprised by the fact that there is no limit.

erizocosmico:
Fixed all the comments @agarciamontoro

sql/cache_test.go
```diff
@@ -0,0 +1,167 @@
...
+		cache := newHistoryCache(mockMemory{}, fixedReporter(51, 50))
+
+		err := cache.Put(1, "foo")
+		require.Error(err)
+		require.True(ErrNoMemoryAvailable.Is(err))
+	})
+
+	t.Run("free required to add entry", func(t *testing.T) {
+		require := require.New(t)
+		var freed bool
```
agarciamontoro: Another unchecked `freed`.

sql/cache_test.go
```diff
@@ -0,0 +1,167 @@
...
+		cache := newRowsCache(mockMemory{}, fixedReporter(51, 50))
+
+		err := cache.Add(Row{1, "foo"})
+		require.Error(err)
+		require.True(ErrNoMemoryAvailable.Is(err))
+	})
+
+	t.Run("free required to add entry", func(t *testing.T) {
+		require := require.New(t)
+		var freed bool
```
agarciamontoro: The last unchecked `freed`.

erizocosmico:
Fixed

sql/cache.go
```diff
@@ -0,0 +1,131 @@
...
+	v, ok := l.cache.Get(k)
+	if !ok {
+		return nil, ErrKeyNotFound.New(k)
+	}
+
+	return v, nil
+}
+
+func (l *lruCache) Free() {
+	l.cache, _ = lru.New(l.size)
```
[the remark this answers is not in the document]
erizocosmico: Yup, only error that can happen is if size is negative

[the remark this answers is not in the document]
juanjux: True. Negative or zero.

sql/memory.go
```diff
@@ -0,0 +1,198 @@
...
+type Freeable interface {
+	// Free the memory.
+	Free()
+}
+
+// KeyValueCache is a cache of key value pairs.
+type KeyValueCache interface {
+	// Put a new value in the cache. If there is no more memory and the cache is
+	// not Freeable it will try to free some memory from other freeable caches.
+	// If there's still no more memory available, it will fail and erase all the
```
ajnavarro: This shouldn't be the case in all the implementations. Per example, a warmed cache can improve a lot consecutive queries execution over more or less the same data.

sql/memory.go
```diff
@@ -0,0 +1,198 @@
...
+func (m *MemoryManager) HasAvailable() bool {
+	return HasAvailableMemory(m.reporter)
+}
+
+// DisposeFunc is a function to completely erase a cache and remove it from the manager.
+type DisposeFunc func()
+
+// NewLRUCache returns an empty LRU cache and a function to dispose it when it's
+// no longer needed.
+func (m *MemoryManager) NewLRUCache(size uint) (KeyValueCache, DisposeFunc) {
```
ajnavarro: This is intended to be used per query or while the SQL engine is running?

ajnavarro: the same question for all the cache implementations

erizocosmico: While it's running.

sql/memory.go
erizocosmico: Ok, will remove that from the doc

PR #804, opened 2019-08-14
Fix typos and make style consistent

agarciamontoro:
- Change all functions descriptions in README.md to be consistent. I decided to stick to the lower-case present-tense style, as each row is easily read as `Function X returns/changes/adds... Y`. This is opinionated (consistency should be not), so I'm ok discussing the best style and changing it again once we agree on one.
- Fix typos in supported_clients.md.

### DIFF
--- a/SUPPORTED_CLIENTS.md
+++ b/SUPPORTED_CLIENTS.md
@@ -1,6 +1,6 @@
 # Supported clients
 
-These are the clients we actively test against to check are compatible with go-mysql-server. Other clients may also work, but we don't check on every build if we remain compatible with them.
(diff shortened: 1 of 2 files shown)


SUPPORTED_CLIENTS.md
```diff
@@ -16,7 +16,7 @@ These are the clients we actively test against to check are compatible with go-m
 - Java/JVM
   - [mariadb-java-client](#mariadb-java-client)
 - Go
-  - [go-mysql-driver/mysql](#go-mysql-driver-mysql)
+  - [go-mysql-driver/mysql](#go-sql-drivermysql)
```
ajnavarro: just to check, is this anchor correct?

agarciamontoro: It is. It did not work before the change.

PR #806, opened 2019-08-16
Check for dead sockets before timeout and enforce timeouts

### LINKED ISSUE #800: Long processing queries don't stop if the connection is dead/killed until the timeout triggers

juanjux:
Fixes #800 at least in Linux (once merged I'll open issues for Darwin and Windows).
Related to src-d/gitbase-spark-connector-enterprise#81

This supersedes #801, the timeout checking part is basically the same with minor changes. What this adds is another goroutine that, using the Linux `/proc` filesystem will check if the socket is closed and cancel the connection and queries in that case.

While this is mergeable, I'll ask to wait until I say it can be merged because I want to try to get rid of the timeout checking goroutine. The current problem is that currently even if there is a timeout, the socket state won't change to `CLOSE_WAIT` until there is a real `read()` or `write()` to it so the timeout checking goroutine is still needed, but I wan't to try to configure keepalives for the connection and see if with this the timeout actually changes the socket state (and thus the connection checking goroutine can reap it). Also I would like to get an Ok from @lwsanty that found a problem in the previous PR.

Bonus: a couple typos fixed in the tests.

cc @lwsanty : please check that this branch works for you with your previously failing test (and if you can link the test so I can maybe add it to the unittests of this project).


server/server.go
```diff
@@ -13,6 +13,8 @@ import (
 // Server is a MySQL server for SQLe engines.
 type Server struct {
 	Listener *mysql.Listener
+	h        *Handler
+	l        *Listener
```
golangcibot: `l` is unused (from `structcheck`)

server/handler.go
```diff
@@ -79,6 +113,8 @@ func (h *Handler) ComQuery(
 	callback func(*sqltypes.Result) error,
 ) (err error) {
 	ctx := h.sm.NewContextWithQuery(c, query)
+	newCtx, cancel := context.WithCancel(ctx)
```
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

server/handler_linux_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require := require.New(t)
+	e := setupMemDB(require)
+
+	port, err := getFreePort()
+	require.NoError(err)
+
+	ready := make(chan struct{})
+	go brokenTestServer(t, ready, port)
+	<-ready
+	conn, err := net.Dial("tcp", "localhost:"+port)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

server/handler_test.go
```diff
@@ -252,3 +211,83 @@ func TestSchemaToFields(t *testing.T) {
...
+func TestOkClosedConnection(t *testing.T) {
+	require := require.New(t)
+	e := setupMemDB(require)
+	port, err := getFreePort()
+	require.NoError(err)
+
+	ready := make(chan struct{})
+	go okTestServer(t, ready, port)
+	<-ready
+	conn, err := net.Dial("tcp", "localhost:"+port)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

server/handler_test.go
```diff
@@ -252,3 +211,83 @@ func TestSchemaToFields(t *testing.T) {
...
+		),
+		0,
+	)
+	h.AddNetConnection(&conn)
+	c2 := newConn(2)
+	h.NewConnection(c2)
+
+	wait := tcpCheckerSleepTime * 2
+	if wait < 4 {
+		wait = 4
```
golangcibot: ineffectual assignment to `wait` (from `ineffassign`)

internal/sockstate/netstat.go
```diff
@@ -0,0 +1,82 @@
...
+// skState type represents socket connection state
+type skState uint8
+
+func (s skState) String() string {
+	return skStates[s]
+}
+
+// Socket states
+const (
+	Established skState = 0x01
```
golangcibot: SA9004: only the first constant in this group has an explicit type (from `staticcheck`)

internal/sockstate/netstat_linux.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+	return tab, br.Err()
+}
+
+// tcpSocks returns a slice of active TCP sockets containing only those
+// elements that satisfy the accept function
+func tcpSocks(accept AcceptFn) ([]sockTabEntry, error) {
+	f, err := os.Open(pathTCPTab)
+	defer f.Close()
```
golangcibot: SA5001: should check returned error before deferring f.Close() (from `staticcheck`)

server/handler_test.go
```diff
@@ -252,3 +211,83 @@ func TestSchemaToFields(t *testing.T) {
...
+	require := require.New(t)
+	e := setupMemDB(require)
+	port, err := getFreePort()
+	require.NoError(err)
+
+	ready := make(chan struct{})
+	go okTestServer(t, ready, port)
+	<-ready
+	conn, err := net.Dial("tcp", "localhost:"+port)
+	defer conn.Close()
```
golangcibot: SA5001: should check returned error before deferring conn.Close() (from `staticcheck`)

server/handler_test_common.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer l.Close()
+	return strconv.Itoa(l.Addr().(*net.TCPAddr).Port), nil
+}
+
+func testServer(t *testing.T, ready chan struct{}, port string, breakConn bool) {
+	l, err := net.Listen("tcp", ":"+port)
+	defer l.Close()
```
golangcibot: SA5001: should check returned error before deferring l.Close() (from `staticcheck`)

server/handler_test_common.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if !breakConn {
+		defer conn.Close()
+
+		_, err = ioutil.ReadAll(conn)
+		if err != nil {
+			t.Fatal(err)
+		}
+	} // else: dirty return without closing or reading to force the socket into TIME_WAIT
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

server/handler.go
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

server/handler_test.go
golangcibot: ineffectual assignment to `wait` (from `ineffassign`)

internal/sockstate/netstat_linux.go
```diff
@@ -0,0 +1,208 @@
+package sockstate
```
erizocosmico: Should have a `// +build linux`

internal/sockstate/netstat_darwin.go
```diff
@@ -0,0 +1,11 @@
+package sockstate
```
erizocosmico: Should have a `// +build darwin`

internal/sockstate/netstat_linux.go
```diff
@@ -0,0 +1,208 @@
...
+	if len(fields) < 2 {
+		return nil, fmt.Errorf("sockstate: not enough fields: %v", s)
+	}
+	var ip net.IP
+	var err error
+	switch len(fields[0]) {
+	case ipv4StrLen:
+		ip, err = parseIPv4(fields[0])
+	default:
+		log.Fatal("Bad formatted string")
```
erizocosmico: we should add more info here about which string was badly formatted

internal/sockstate/netstat_linux.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	return tab, br.Err()
+}
+
+// tcpSocks returns a slice of active TCP sockets containing only those
+// elements that satisfy the accept function
+func tcpSocks(accept AcceptFn) ([]sockTabEntry, error) {
+	f, err := os.Open(pathTCPTab)
+	defer func() {
+		_ = f.Close()
```
erizocosmico: if you're ignoring the error of f.Close() then just `defer f.Close()`

internal/sockstate/netstat_windows.go
```diff
@@ -0,0 +1,11 @@
+package sockstate
```
erizocosmico: should have `// +build windows`

internal/sockstate/sockstate.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	switch len(socks) {
+	case 0:
+		return Finished, nil
+	case 1:
+		if socks[0].State == CloseWait {
+			return Broken, nil
+		} else {
+			return Other, nil
```
erizocosmico: else branch not needed, previous branch already returned

internal/sockstate/netstat_darwin.go
juanjux: AFAIK naming the module as mod_OSNAME.go has the same effect (at least Goland and the Go toolchain recognise it). Am I wrong?

internal/sockstate/netstat_linux.go
juanjux: Ditto.

internal/sockstate/netstat_linux.go
juanjux: It's so GolangCI shut ups, which it doesn't which just a `defer close()`

internal/sockstate/netstat_windows.go
juanjux: Ditto.

internal/sockstate/netstat_darwin.go
juanjux: PS: added anyway, explicit is better.

internal/sockstate/sockstate.go
juanjux: True. As a somewhat pedantic note, it's usually considered good practice to add an else branch even when not needed but I also prefer to remove unnecessary branches, lines of code and indentation when possible, so done.

erizocosmico: In Go that's not considered a good practice. Golang linters complain about it

juanjux: Well, that doesn't suprise me :rofl:

server/handler.go
```diff
@@ -96,13 +133,104 @@ func (h *Handler) ComQuery(
...
 	var proccesedAtLeastOneBatch bool
+
+	// Reads rows from the row reading goroutine
+	rowChan := make(chan sql.Row)
+	// To send errors from the two goroutines to the main one
+	errChan := make(chan error)
+	// To close the goroutines
+	quit := make(chan struct{})
+
+	// Default waitTime is one 1 minute if there is not timeout configured, in which case
```
agarciamontoro: Typo: should be either "one" or "1".

juanjux:
Yesterday I tried many things to be able to get rid of the timeout-enforcing goroutine, without success. I'll try a couple things more today and change the status to mergeable anyway. Windows support will come in another PR so I don't stall this too much, since it's very needed.

server/handler.go
juanjux: Fixed, thanks!

juanjux:
Ok, so tried keepalives (higher Go level and lower level using syscalls), deadlines, and many other things. Looks like we can't get rid of the second goroutine so since this works this could be merged when reviews pass. PTAL @erizocosmico @ajnavarro.

I'll work on adding Windows support for the dead socket detector on another PR.

server/context.go
```diff
@@ -2,10 +2,9 @@ package server
 
 import (
 	"context"
-	"sync"
-
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/src-d/go-mysql-server/sql"
+	"sync"
```
ajnavarro: these imports are disordered, right?

server/handler_linux_test.go
```diff
@@ -0,0 +1,53 @@
+package server
+
+import (
+	"fmt"
+	"github.com/opentracing/opentracing-go"
+	"github.com/src-d/go-mysql-server/sql"
+	"github.com/stretchr/testify/require"
+	"net"
```
ajnavarro: these imports are unordered. check format.

juanjux:
@ajnavarro gofmt ran on the entire project and goimports on the modified directories.

PR #812, opened 2019-09-04
Removed pilosa-dependent files from Windows compilation

Hydrocharged:
[Pilosa](https://github.com/pilosa/pilosa/) makes use of the following call which does not work on Windows:
https://github.com/pilosa/pilosa/blob/bc9747cc0f19702d9753de7ea9375d8311dfc706/syswrap/mmap.go#L53
This causes any files that refer to Pilosa to fail compilation, which in turn causes builds for this project to also fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ CREATE INDEX foo ON table USING driverid (col1, col2) WITH (async = true)
 
 ### Old `pilosalib` driver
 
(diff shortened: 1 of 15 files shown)


Hydrocharged:
I submitted an issue to Pilosa regarding Windows not building on their project, and it was stated that there is no plan to support Windows natively.

https://github.com/pilosa/pilosa/issues/2059#issuecomment-528390505

PR #822, opened 2019-09-16
Implemented DELETE

Hydrocharged:
Implemented DELETE functionality, along with tests.

PR #825, opened 2019-09-24
Implemented REPLACE

Hydrocharged:
This PR is a follow-up to https://github.com/src-d/go-mysql-server/pull/822 and includes that commit's changes as well.

Looking over the [MySQL REPLACE documentation](https://dev.mysql.com/doc/refman/8.0/en/replace.html), I see that it mentions a REPLACE being either a DELETE then INSERT, or just an INSERT. I figured that there were two ways to implement this: as another interface with a unique function (`Replacer` with a `Replace` function for example), or by leveraging the already-included `Insert` and `Delete` functions from their respective interfaces. I decided to go with the latter option for the below reasons.

One point is in the reduction of code needed for someone wanting to implement the library. The more that the framework handles (without appreciable loss of freedom), the better. Then there is also the consideration of triggers. If DELETE and INSERT calls should make use of triggers (which are probably already built into the implementing `Delete` and `Insert` functions), then simply calling those _as normal_ will allow for correct behavior by default.

The biggest downside that I could see is that `Delete` is now forced to check if the given row exists, and return the appropriate error if not. For some implementations, this will not make an impact on the written code nor performance, but for others it definitely will. I felt that the trade off was worth it, but I'm interesting in hearing your opinions as well.

A third potential option is to allow for a custom `Replace` function if provided, otherwise defaulting to DELETE + INSERT from earlier. For example:
```go
type Replacer interface {
	Deleter
	Inserter
}

type CustomReplacer interface {
	Replace(*Context, Row) error
}
```
This would solve both of the problems as far as I can tell, but would not follow in the pattern established for every other keyword, and thus I decided against this.

### DIFF
--- a/sql/parse/parse.go
+++ b/sql/parse/parse.go
@@ -148,6 +148,8 @@ func convert(ctx *sql.Context, stmt sqlparser.Statement, query string) (sql.Node
 		return convertUse(n)
 	case *sqlparser.Rollback:
 		return plan.NewRollback(), nil
+	case *sqlparser.Delete:
+		return convertDelete(ctx, n)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


sql/plan/insert.go
```diff
@@ -119,17 +137,37 @@ func (p *InsertInto) Execute(ctx *sql.Context) (int, error) {
...
 		}
 
+		if replaceable != nil {
+			if err = replaceable.Delete(ctx, row); err != nil {
+				if err != sql.ErrDeleteRowNotFound {
+					_ = iter.Close()
+					return i, err
+				}
+			} else {
+				i++
```
agarciamontoro: Do we understand the whole DELETE+INSERT as two different updates? Internally they are, but I think that if we count the UPDATE statements as just one (do we?) then we should do the same thing here. WDYT, @erizocosmico?

sql/parse/parse.go
```diff
@@ -373,6 +373,11 @@ func convertInsert(ctx *sql.Context, i *sqlparser.Insert) (sql.Node, error) {
 		return nil, ErrUnsupportedSyntax.New(i)
 	}
 
+	isReplace := false
+	if i.Action == sqlparser.ReplaceStr {
+		isReplace = true
+	}
```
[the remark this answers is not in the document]
Hydrocharged: It is definitely is, thank you for catching that!

sql/plan/insert.go
Hydrocharged: From looking at the [documentation](), it looks like REPLACE should distinctly act as a shortcut to the two separate calls. The returned count should mimic this behavior:

> The REPLACE statement returns a count to indicate the number of rows affected. This is the sum of the rows deleted and inserted. If the count is 1 for a single-row REPLACE, a row was inserted and no rows were deleted.

I wish I could find it again (my Googling has failed me), but I also remember reading about REPLACE causing a user problems because it was triggering a CASCADE DELETE and messing up a lot more of their data, and the recommendations to that thread was to use UPDATE instead since it's not doing two operations. Totally anecdotal, but it also seems to suggest that this should be treated explicitly as the two separate operations.

agarciamontoro: Skipped that part of the documentation, you are absolutely right. Thank you!

PR #827, opened 2019-09-30
plan: return types in lowercase in SHOW CREATE TABLE

erizocosmico:
Fixes https://github.com/src-d/gitbase/issues/962

It solves a compatibility issue with SQLAlchemy, as it did not return
the types the same way MySQL does.


ajnavarro:
Seems like nodejs integration tests are failing.

ajnavarro:
Ignoring node integration tests for now and merging.

PR #831, opened 2019-10-01
Fixed null errors during value comparisons

Hydrocharged:
This bug was discovered while creating https://github.com/src-d/go-mysql-server/pull/832, and is required for one of the tests in that PR to pass.

Null values weren't being handled in most of the value comparison code, and was causing errors for valid queries. This PR fixes those issues.


ajnavarro:
Ignoring node integration tests by now and merging.

PR #832, opened 2019-10-01
Implemented UPDATE

Hydrocharged:
Notice: one of the tests will fail until https://github.com/src-d/go-mysql-server/pull/831 is merged in, as it is required to fix one of the bugs that that test catches.

This is an implementation of UPDATE, which brings with it a new interface. Unlike the other interfaces, this one supplies two rows, the original row, and the updated row. This decision was made so that implementations that rely on primary keys can have access to the old primary key if need be.

~Besides the big change, I've also caught a bug where supplying a non-existent column to ORDER BY will cause the query to still succeed, while ignoring the ORDER BY clause. Testing against MySQL showed that an error is the proper behavior, and this has been corrected. Since the change was small, I added it here rather than making an entirely-new PR for it.~ This was fixed by https://github.com/src-d/go-mysql-server/pull/818.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -120,7 +120,7 @@ func (e *Engine) Query(
 	case *plan.CreateIndex:
 		typ = sql.CreateIndexProcess
 		perm = auth.ReadPerm | auth.WritePerm
-	case *plan.InsertInto, *plan.DeleteFrom, *plan.DropIndex, *plan.UnlockTables, *plan.LockTables:
+	case *plan.InsertInto, *plan.DeleteFrom, *plan.Update, *plan.DropIndex, *plan.UnlockTables, *plan.LockTables:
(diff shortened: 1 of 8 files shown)


ajnavarro:
@Hydrocharged #831 merged! you can rebase when you wish. Thanks!

Hydrocharged:
@ajnavarro It has been rebased!

agarciamontoro:
Sorry for the delay on this one, @Hydrocharged. When rebasing, build will fail because of https://github.com/src-d/go-mysql-server/pull/835. You would probably just need to pass a `*sql.Context` wherever it is needed.

Hydrocharged:
The `*sql.Context`s have been passed around! 😄

erizocosmico:
Thank you for the contribution!

PR #840, opened 2019-10-11
Fix integer literals parsing

### LINKED ISSUE #834: Returned types do not fit the schema
All integer types (except for `uint64`) are converted to `int64`, regardless of what was defined in the original schema of the table. This creates an overhead in space that should not be there. It also creates weird casting bugs in very particular scenarios (e.g., gitbase connected to Spark).

## Example

The `typestable` table created in [`engine_test.go#L2674-L2691`](https://github.com/src-d/go-mysql-server/blob/b09e8c1c8a/engine_test.go#L2674-L2691) has columns of each and every type of signed and unsigned integers.

However, as soon as we make an INSERT into it and try to SELECT all the columns from it, as shown [in one of the tests](https://github.com/src-d/go-mysql-server/blob/b09e8c1c8a/engine_test.go#L1858-L1873), the integer data types are all converted to `int64`, regardless of what they were originally (`uint8`, `int16`, `uint32`...). The only exception is `uint64`, whose maximum is of course greater than the maximum of `int64`.

Actually, these tests are wrong as well, as the returned value should not always be converted to `int64` but to the type specified in the schema.

#### Parse the literals as the smallest integer possible

The integer literals are always parsed as `int64`, defaulting to `uint64` if the first parsing fails. See [sql/parse/parse.go#L945-L956](https://github.com/src-d/go-mysql-server/blob/b09e8c1c8a/sql/parse/parse.go#L945-L956).

As the `TODO` comment there, I propose to parse the literal to the smallest integer possible, defaulting to greater integers if necessary, as follows:
```
int8 < uint8 < int16 < uint16 < int32 < uint32 < int64 < uint64
```
An example of such a change is in [my fork](https://github.com/agarciamontoro/go-mysql-server/commit/922906c52673766e435b764e93d558c5b93e6ed2), although it was a test that considered only `int32`, so it should be expanded to all the other types.

#### Convert to greater types when necessary

The smallest integer possible for a parsed literal my not be the expected type for it, so we should convert them back to greater types when needed. 

For example, when doing the INSERT queries, a literal may be parsed as an `int8` when the column type is `int64`. The conversion will be safe as the parsed number will always be smaller or equal than the maximum accepted by the type (assuming the query is correct; the user can write a number greater than the maximum allowed by the column, but failing there is the correct thing to do).

For this case of the INSERT, I propose to implement a new rule in the plan that converts the parsed literals into the right type specified by the column.

This would probably be a big change, as there are some hard-coded literals with specific types that will fail after applying the previous solution (there are [some](https://github.com/agarciamontoro/go-mysql-server/commit/d2e6f4b6f109a4cf62fda2bef65fe75933287dcf) [examples](https://github.com/agarciamontoro/go-mysql-server/commit/93a8634fcc653ca77e78d58a2e5b930e557de164) [in](https://github.com/agarciamontoro/go-mysql-server/commit/d239aa4d9eb363016a8cedac2ac2b74d55761ba6) [my](https://github.com/agarciamontoro/go-mysql-server/commit/29a2f812d346bcc62f0e041f6546ef0f8c24b160) [branch](https://github.com/agarciamontoro/go-mysql-server/commit/02087f4417e9e6aa0819e699d584ce14b053e9a3)). Fortunately, the test suite is quite comprehensive and covers most corner cases

**Note**: as this is probably a big change, as described before, I would like to discuss the details with @src-d/data-processing before going down that road. What do you think?

agarciamontoro:
This PR fixes #834 with some changes:

- Modify `convertInt` so it returns the smallest representation allowed. See https://github.com/src-d/go-mysql-server/commit/acb2bc0f3bdae0ab853d3360bda24f0d7f11fea4.
- ~Add a rule to convert integer literals to the specified types in the schema in INSERT nodes. See https://github.com/src-d/go-mysql-server/commit/2eb11bbb08f3cc3c5b07247ba1c1e91032667e7e.~ After discussion with @erizocosmico (check https://github.com/src-d/go-mysql-server/pull/840#discussion_r333931390), I've completely removed the integer conversion rule and moved its logic to Insert.Execute function. Much less code, which is always better. See instead https://github.com/src-d/go-mysql-server/pull/840/commits/653fdde5b1d0d0a61c1931d187b1909b30571b44#diff-7e21a9fc108efea330a88d73d2a81023.
- Adapt several parts of the code that broke after the above changes, namely:
    - `getInt64Literal` function was modified so it converts any integer to int64. See https://github.com/src-d/go-mysql-server/commit/10d13fc633f6afc0a6a5214d3516cd2bc06beb74.
    - an explicit conversion to int64 was added in GROUP BY parsing, so any integer that fits is allowed. See https://github.com/src-d/go-mysql-server/commit/2320b073305611ddd8dc5e2a74291c1e00de5b22.
    - the ROUND and YEARWEEK functions were also modified so they can work with all types of integers. See https://github.com/src-d/go-mysql-server/commit/97a23158aa8444cf25171100c6baf7c6537124b9
    - most of the changes in this PR are in tests, as they were tailored to the previous int64-only design. These changes make the tests expect the right types. Review with care here, please :pray: : https://github.com/src-d/go-mysql-server/commit/00a883389ab70381cfd03341d5fb1b2913830285
    - some more boilerplate changes in pilosa code so it handles 8 and 16 bits signed and unsigned integers values as well. See https://github.com/src-d/go-mysql-server/commit/5c38a0ddcf90f5c197effdeb796439520839e33d
    - I also needed to modify the conversion back to `sqltypes.Value` in `numberT.SQL` function, as it failed when 8 and 16 bits signed and unsigned integers values were passed, resulting in empty rows when doing, for example, `SELECT 1`. See https://github.com/src-d/go-mysql-server/commit/ffae1d2959be5b6ae6edec0006e5045a8f01f277

Maybe the most comfortable way to review this is:
- Check first the new `convertInt` in https://github.com/src-d/go-mysql-server/commit/acb2bc0f3bdae0ab853d3360bda24f0d7f11fea4 (if you know a way of making that code a bit more beautiful, I'd be glad to hear it)
- Check then ~the new rule for integer conversion https://github.com/src-d/go-mysql-server/commit/2eb11bbb08f3cc3c5b07247ba1c1e91032667e7e and its test https://github.com/src-d/go-mysql-server/commit/22600f9252fb1a24b33b945c76952b2437466eed~ the modified `Insert.Execute` function: https://github.com/src-d/go-mysql-server/pull/840/commits/653fdde5b1d0d0a61c1931d187b1909b30571b44#diff-7e21a9fc108efea330a88d73d2a81023
- Review the minor modifications: https://github.com/src-d/go-mysql-server/commit/10d13fc633f6afc0a6a5214d3516cd2bc06beb74, https://github.com/src-d/go-mysql-server/commit/2320b073305611ddd8dc5e2a74291c1e00de5b22, https://github.com/src-d/go-mysql-server/commit/97a23158aa8444cf25171100c6baf7c6537124b9, https://github.com/src-d/go-mysql-server/commit/5c38a0ddcf90f5c197effdeb796439520839e33d and https://github.com/src-d/go-mysql-server/commit/ffae1d2959be5b6ae6edec0006e5045a8f01f277
- Review the changes in existing tests, which will be tedious: https://github.com/src-d/go-mysql-server/commit/00a883389ab70381cfd03341d5fb1b2913830285 and the new additions: https://github.com/src-d/go-mysql-server/pull/840/commits/03f9766afc6eca5baca5ee96ddaa24a29c2be292, https://github.com/src-d/go-mysql-server/pull/840/commits/2917915a7fd02c4e0ababb8021e7bb27c15c8f08, https://github.com/src-d/go-mysql-server/pull/840/commits/cc99d6c45e707413e5dcf73f876de20d30131032 and https://github.com/src-d/go-mysql-server/pull/840/commits/fe3a84c1343193eed1889c8b49b1aa39954d1aed.

This one is my first big contribution to go-mysql-server, so please review and test thoroughly. Comments and suggestions are more than welcome.


sql/analyzer/resolve_insert_literals.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+var errWrongNumberOfValues = errors.NewKind("the number of values to insert differ from the expected columns")
+
+func convertIntegerLiteralsInsert(ctx *sql.Context, analyzer *Analyzer, originalNode sql.Node) (sql.Node, error) {
+  span, _ := ctx.Span("resolve_insert_literals")
+  defer span.Finish()
+
+  return plan.TransformUp(originalNode, func(node sql.Node) (sql.Node, error) {
+
```
erizocosmico: leftover blank line

sql/parse/parse.go
```diff
@@ -938,22 +944,51 @@ func isAggregateFunc(v *sqlparser.FuncExpr) bool {
 	return v.IsAggregate()
 }
 
+// Convert an integer, represented by the specified string in the specified
+// base, to its smallest representation possible, out of:
+// int8, uint8, int16, uint16, int32, uint32, int64 and uint64
+func convertInt(value string, base int) (sql.Expression, error) {
+	i8, err := strconv.ParseInt(value, base, 8)
```
erizocosmico: Instead of the great pyramid of Ghiza what do you think of doing this to keep it a little bit more flat? 😛 

```go
if i8, err := stconv.ParseInt(value, base, 8); err == nil {
    return expression.NewLiteral(int8(i8), sql.Int8), nil
}

if ui8, err := stconv.ParseUint(value, base, 8); err == nil {
    return expression.NewLiteral(unt8(ui8), sql.Uint8), nil
}

// the rest of the checks

ui64, err := strconv.ParseUint(value, base, 64)
if err != nil {
    return nil, err
}

return expression.NewLiteral(uint64(ui64), sql.Uint64), nil
```

sql/analyzer/resolve_insert_literals.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/src-d/go-mysql-server/sql"
+  "github.com/src-d/go-mysql-server/sql/expression"
+	"github.com/src-d/go-mysql-server/sql/plan"
+  errors "gopkg.in/src-d/go-errors.v1"
+
+)
+
+var errWrongNumberOfValues = errors.NewKind("the number of values to insert differ from the expected columns")
+
+func convertIntegerLiteralsInsert(ctx *sql.Context, analyzer *Analyzer, originalNode sql.Node) (sql.Node, error) {
```
erizocosmico: I think we should let the table implementations handle this. When the literals are parsed, they are converted to the minimum type of int possible, then the table is in charge of converting it.

If we really want to make this transparent for table implementations, then I'd go for doing `column.Type.Convert(theLiteralValue)` inside the `InsertInto` execution method, but I would not add a new rule in the analyzer for this specific case.

WDYT @src-d/data-processing?

sql/parse/parse.go
agarciamontoro: This is what I wanted to hear, that arrow shape was ugly. Thanks!

agarciamontoro: Done in https://github.com/src-d/go-mysql-server/pull/840/commits/a7aa47f53e96d9088894e3f9efe03f6e2ac4ef83

sql/parse/parse.go
```diff
@@ -948,39 +948,35 @@ func isAggregateFunc(v *sqlparser.FuncExpr) bool {
...
-					}
-					return expression.NewLiteral(int32(i32), sql.Int32), nil
-				}
-				return expression.NewLiteral(uint16(ui16), sql.Uint16), nil
-			}
-			return expression.NewLiteral(int16(i16), sql.Int16), nil
-		}
-		return expression.NewLiteral(uint8(ui8), sql.Uint16), nil
+	if i8, err := strconv.ParseInt(value, base, 8); err == nil {
+    return expression.NewLiteral(int8(i8), sql.Int8), nil
```
erizocosmico: i think your editor did not run gofmt to this part

agarciamontoro: :thinking: It looks good on my side, I'll check

agarciamontoro: Spaces instead of tabs, ofc. Force pushed as https://github.com/src-d/go-mysql-server/pull/840/commits/3034d2d03a9d2a485470e630873fa40b9d87f61e

sql/analyzer/resolve_insert_literals.go
agarciamontoro: I think that the conversion should be done by us. If we convert string literals to proper dates, we should do the same here. I'd argue that `int8` is as different from `uint64` as a date is different from a string.

But maybe it makes sense to do it inside InsertInto, not as a new rule. I'll wait for the rest of opinions.

agarciamontoro:
I'm just noticing that I did not run gofmt over the new files. I'll fix the formatting in the next commit.

sql/analyzer/resolve_insert_literals.go
agarciamontoro: Done in https://github.com/src-d/go-mysql-server/pull/840/commits/6d4af9b32ba3d74f35ab0d518029332cb797ceac

agarciamontoro:
As per @erizocosmico's comment (check https://github.com/src-d/go-mysql-server/pull/840#discussion_r333931390), I've completely removed the integer conversion rule and moved its logic to Insert.Execute function. Much less code, which is always better.

I've added it as a new commit (https://github.com/src-d/go-mysql-server/pull/840/commits/653fdde5b1d0d0a61c1931d187b1909b30571b44) for if someone still prefers the other alternative, but I can rebase it before merging.

sql/analyzer/resolve_insert_literals.go
agarciamontoro: Solved in https://github.com/src-d/go-mysql-server/commit/653fdde5b1d0d0a61c1931d187b1909b30571b44.

agarciamontoro:
The latest four commits add missing cases to tests of modified functions. I think that CodeCoverage should be happy now.

erizocosmico:
Can you rebase? there are some conflicts nowl

agarciamontoro:
Rebased. Codecov is still complaining, though. I'll take a look.

go.mod
```diff
@@ -26,3 +26,5 @@ require (
 	gopkg.in/yaml.v2 v2.2.2
 	vitess.io/vitess v3.0.0-rc.3.0.20190602171040-12bfde34629c+incompatible
 )
+
+go 1.13
```
carlosms: My understanding is that this allows the code to use go 1.13 features, but we are supposed to support the last 2 stable versions. Should we set this to 1.12 instead?

agarciamontoro: Oops, my bad. I use go 1.13 by default and it changes the `go.mod` automatically, so I guess I did a `git add -a` and included it. I'll remove it. Thanks for the catch!

agarciamontoro:
Force pushed without changes in `go.mod`. Thanks for the catch, @carlosms!

PR #843, opened 2019-10-17
plan: change how cancellation is handled in exchanges

erizocosmico:


lwsanty:
as I told before this change does not affect https://github.com/src-d/gitbase/issues/955#issuecomment-542249137 but at the same time does not break anything afaik

PR #852, opened 2019-10-22
ci: add python-mysql and python-sqlalchemy to matrix

### LINKED ISSUE #846: 2 integration tests are not included in travis
We have integration tests for `python-mysql`, `python-pymysql`, and `python-sqlalchemy`.
In `.travis.yml` we are only executing `python-pymysql`.

On a related note, the test for sqlalchemy is using the pymysql driver. Since Superset uses sqlalchemy with the mysqlclient driver, maybe we could replace it to be more aligned with our stack.

erizocosmico:
Closes #846


erizocosmico:
Fixed

agarciamontoro:
@erizocosmico great! What do you think about adding `test-server` to `.gitignore` so it does not happen again?

erizocosmico:
That happened because I ran `make integration` locally, but that's not usually the flow for development, only for CI. I can add it, though

erizocosmico:
Added

agarciamontoro:
> That happened because I ran make integration locally, but that's not usually the flow for development, only for CI. I can add it, though

Yeah, but I have tried exactly that some times and ended up with the binary in my local git history by mistake.

> Added

Thanks!

PR #854, opened 2019-10-22
sql: finish root span of the query

### LINKED ISSUE #853: Root opentracing span is not finished
The root opentracing span (https://github.com/src-d/go-mysql-server/blob/6f000e20fb418cf46eda4025c11ad26c5f0e4dc9/server/context.go#L97) which is created when serving queries via the SQL engine is never finished. When using jaeger, this results in traces with a name of `<trace-without-root-span>`.

erizocosmico:
Fixes #853

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
+
 	sqle "github.com/src-d/go-mysql-server"
 	"github.com/src-d/go-mysql-server/auth"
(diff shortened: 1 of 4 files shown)


erizocosmico:
@agarciamontoro yup. But it's only whitespace changes after all, so doesn't matter much. Some commit might have gone to master without gofmt

erizocosmico:
Rebased

erizocosmico:
CI is failing because of a random C# issue that happens from time to time downloading and setting up dotnet

PR #855, opened 2019-10-23
Add progress for each partition in SHOW PROCESSLIST

carlosms:
This is a proposal to add some more information to the `SHOW PROCESSLIST` output.
It adds a counter for each active partition. A partition gets removed from the list when it finishes.

This is how it looks:
```
$ mysql -u root -h 127.0.0.1 -P 3306 gitbase --execute "show processlist"
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------+-----------------------+
| Id   | User | Host            | db      | Command | Time | State                                                            | Info                  |
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------+-----------------------+
|  395 | root | 127.0.0.1:35250 | gitbase | query   |    2 | cangallo(45/?), commits(0/3), octoprint-tft(45/?), upsilon(45/?) | select * from commits |
|  396 | root | 127.0.0.1:35250 | gitbase | query   |    0 | running                                                          | show processlist      |
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------+-----------------------+

$ mysql -u root -h 127.0.0.1 -P 3306 gitbase --execute "show processlist"
+------+------+-----------------+---------+---------+------+------------------------------+-----------------------+
| Id   | User | Host            | db      | Command | Time | State                        | Info                  |
+------+------+-----------------+---------+---------+------+------------------------------+-----------------------+
|  395 | root | 127.0.0.1:35256 | gitbase | query   |   11 | commits(2/3), upsilon(229/?) | select * from commits |
|  398 | root | 127.0.0.1:35256 | gitbase | query   |    0 | running                      | show processlist      |
+------+------+-----------------+---------+---------+------+------------------------------+-----------------------+
```
```
$ mysql -u root -h 127.0.0.1 -P 3306 gitbase --execute "show processlist"
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------+
| Id   | User | Host            | db      | Command | Time | State                                                                                                      | Info                                                               |
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------+
|  173 | root | 127.0.0.1:35236 | gitbase | query   |   65 | SquashedTable(commits, commit_blobs, blobs)(0/3), cangallo(1289/?), octoprint-tft(1280/?), upsilon(1281/?) | select * from commits natural join commit_blobs natural join blobs |
|  393 | root | 127.0.0.1:35236 | gitbase | query   |    0 | running                                                                                                    | show processlist                                                   |
+------+------+-----------------+---------+---------+------+------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------+
```
```
$ mysql -u root -h 127.0.0.1 -P 3306 gitbase --execute "show processlist"
+------+------+-----------------+---------+--------------+------+------------------------------+---------------------------------------------------------------------------------------------------------+
| Id   | User | Host            | db      | Command      | Time | State                        | Info                                                                                                    |
+------+------+-----------------+---------+--------------+------+------------------------------+---------------------------------------------------------------------------------------------------------+
|  395 | root | 127.0.0.1:35270 | gitbase | create_index |    8 | commits(2/3), upsilon(160/?) | CREATE INDEX commits_hash_idx ON commits USING pilosa (repository_id, commit_hash) WITH (async = false) |
|  402 | root | 127.0.0.1:35270 | gitbase | query        |    0 | running                      | show processlist                                                                                        |
+------+------+-----------------+---------+--------------+------+------------------------------+---------------------------------------------------------------------------------------------------------+
```

The new output can be used to notice when for some reason a repository takes more time than others to finish.

Because there aren't totals, the counter of each partition cannot be used to get an estimate of the progress made. Right now its only function is to let the user know that the iterator is working.

I am concerned with the `onNext` function, because I'm not sure if updating the progress for every row will have an impact on performance. I'm working on testing this, but so far `gitbase-regression` has been giving me unreliable times.

If we removed `onNext` we could still benefit from having the name of the active partitions, maybe with a `(?/?)` progress.

Another generic improvement would be to have a way to get the total number of rows for a table. Maybe implementing a new interface. I didn't explore this possibility because as far as I know in gitbase we won't be able to use this. But it may be an interesting generic addition later for go-mysql-server itself.

### DIFF
--- a/sql/analyzer/process.go
+++ b/sql/analyzer/process.go
@@ -44,16 +44,25 @@ func trackProcess(ctx *sql.Context, a *Analyzer, n sql.Node) (sql.Node, error) {
 
 			seen[name] = struct{}{}
 
-			notify := func() {
+			onPartitionDone := func(partitionName string) {
 				processList.UpdateProgress(ctx.Pid(), name, 1)
+				processList.RemoveProgressItem(ctx.Pid(), partitionName)
+			}
+
+			onPartitionStart := func(partitionName string) {
+				processList.AddProgressItem(ctx.Pid(), partitionName, -1)
+			}
+
+			onRowNext := func(partitionName string) {
+				processList.UpdateProgress(ctx.Pid(), partitionName, 1)
 			}
 
 			var t sql.Table
 			switch table := n.Table.(type) {
 			case sql.IndexableTable:
-				t = plan.NewProcessIndexableTable(table, notify)
+				t = plan.NewProcessIndexableTable(table, onPartitionDone, onPartitionStart, onRowNext)
 			default:
-				t = plan.NewProcessTable(table, notify)
+				t = plan.NewProcessTable(table, onPartitionDone, onPartitionStart, onRowNext)
 			}
 
 			return plan.NewResolvedTable(t), nil
--- a/sql/plan/process.go
+++ b/sql/plan/process.go
@@ -37,7 +37,7 @@ func (p *QueryProcess) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 		return nil, err
 	}
 
-	return &trackedRowIter{iter, p.Notify}, nil
+	return &trackedRowIter{iter: iter, onDone: p.Notify}, nil
 }
 
 func (p *QueryProcess) String() string { return p.Child.String() }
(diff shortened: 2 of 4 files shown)


erizocosmico:
Overall the functionality and the implementation look good to me, but I just have one doubt: what is the expected way of using this in a client (i.e. gitbase)?

I see ProcessTable having the callbacks, but the client is not supposed to touch ProcessTable at all. In fact, it's a node for internal use and the user should never have to even know it exists.

PR #860, opened 2019-10-29
Add support for DROP VIEW

agarciamontoro:
This PR adds support for `DROP VIEW` statements, as in

```sql
DROP VIEW [IF EXISTS] [db_name1.]view_name1 [, [db_name2.]view_name2, ...] [RESTRICT] [CASCADE]
```

As each view in the statement may come from a different database, I have decided to model the plan as follows:

- There is a parent node, `DropView`, that just contains a list of children of type `SingleDropView`.
- The nodes `SingleDropView` implement the `Databaser` interface so the analyzer can resolve the database in each of them.

The actual removal of the views happen in the `RowIter` from `DropView`, as we need to:

1. Retrieve the ViewRegistry mutex.
2. Check that all views in the children member of `DropView` can be deleted (i.e., they exist or they do not exist but the user has added the `IF EXISTS` keyword to the statement)
3. Delete all of the views listed if the previous step raised no error.
4. Release the mutex.

I first started implementing the removal in the `SingleDropView` nodes, but then it was difficult to check that all views could be deleted and then delete them, all in an atomic operation. The final implementation makes this easy with the new `DeleteList` from `ViewRegistry`.

### DIFF
--- a/sql/parse/util.go
+++ b/sql/parse/util.go
@@ -516,3 +516,56 @@ func maybeList(opening, separator, closing rune, list *[]string) parseFunc {
 		}
 	}
 }
+
+type QualifiedName struct {
+	db   string
+	name string
+}
+
+func readQualifiedIdentifierList(list *[]QualifiedName) parseFunc {
+	return func(rd *bufio.Reader) error {
+		for {
+			var newItem []string
+			err := parseFuncs{
+				skipSpaces,
+				readIdentList('.', &newItem),
+				skipSpaces,
+			}.exec(rd)
+
+			if err != nil {
+				return err
+			}
+
+			if len(newItem) < 1 || len(newItem) > 2 {
+				return errUnexpectedSyntax.New("[db_name.]viewName", strings.Join(newItem, "."))
+			}
+
+			var db, name string
+
+			if len(newItem) == 1 {
+				db = ""
+				name = newItem[0]
+			} else {
+				db = newItem[0]
+				name = newItem[1]
+			}
+
+			*list = append(*list, QualifiedName{db, name})
+
+			r, _, err := rd.ReadRune()
+			if err != nil {
+				if err == io.EOF {
+					return nil
+				}
+				return err
+			}
+
+			switch r {
+			case ',':
+				continue
+			default:
+				return rd.UnreadRune()
+			}
+		}
+	}
+}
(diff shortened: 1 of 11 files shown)


sql/parse/util.go
```diff
@@ -516,3 +516,65 @@ func maybeList(opening, separator, closing rune, list *[]string) parseFunc {
 		}
 	}
 }
+
+// A QualifiedName represents an identifier of type "db_name.table_name"
+type QualifiedName struct {
```
erizocosmico: is this used outside this package? If not, I would keep it private

sql/plan/drop_view.go
```diff
@@ -0,0 +1,148 @@
...
+// node to fail if any of the views in children does not exist.
+type DropView struct {
+	children []sql.Node
+	Catalog  *sql.Catalog
+	ifExists bool
+}
+
+// NewDropView creates a DropView node with the specified parameters,
+// setting its catalog to nil.
+func NewDropView(children []sql.Node, ifExists bool) *DropView {
```
erizocosmico: Instead of having SingleDropView and DropView, what do you think about just keeping DropView and a single drop view would just be a DropView with a single children?

sql/parse/util.go
agarciamontoro: I wanted to keep it private too, but I use it in `views.go` when calling `readQualifiedIdentifierList`.

sql/plan/drop_view.go
agarciamontoro: I don't know if I'm following you :thinking: The `SingleDropView` is just a dummy node containing the view to drop, and I need to have it as one entity so every droppable view can contain its own database to resolve. But then I need them all together in a single node so I can call `DeleteList` and make sure that I check the errors and perform the actual removal in one atomic operation.

sql/parse/util.go
agarciamontoro: Done in https://github.com/src-d/go-mysql-server/pull/860/commits/9b8ecfaa7946a9b87e74946736b6566c205f6f04.

sql/plan/drop_view.go
erizocosmico: Oh, nevermind then, I did not understood correctly what it was doing at first glance

agarciamontoro: @erizocosmico maybe the names are not the best ones, I understand the confusion.

PR #864, opened 2019-11-06
Fix a bug in index assignment that led to incorrect query results, and add index driver to engine_test

zachmu:
In testing out indexing capability with our product's custom IndexDriver, I found a bug. The following query would incorrectly return only a single result on an indexed column for an IndexLookup that didn't implement the Mergeable interface:

SELECT col1 from table where col1 = 1 or col1 = 2;

When the two index lookups couldn't be merged, the first one was returned as the sole lookup for that table. This lookup then got pushed down to the table by the analyzer.

Verifying that this bug was actually fixed ended up being pretty difficult, because engine_test doesn't use an IndexDriver. So I moved the test index types out of assign_indexes_test.go into the memory package, then implemented the Values() method for them so that they could return matching rows from the memory tables. Then I implemented an IndexDriver for the memory package which can be seeded by hand for tests. Finally, I changed TestQueries to test every combination of partitions, index driver, and parallelism, for a total of 12 runs.

The biggest change here is how MergedIndexLookups are handled in the assign_index_tests.go file. What I have now is (I think) more understandable and correct than the previous method of tracking unions and intersections, but did substantially increase the amount of boilerplate in the test code.

Finally, I fixed an unrelated bug in time_test.go, which would fail in my timezone after 4pm due to timezone differences in date calculation.

### DIFF
--- a/sql/analyzer/assign_indexes_test.go
+++ b/sql/analyzer/assign_indexes_test.go
@@ -166,7 +166,7 @@ func TestAssignIndexes(t *testing.T) {
 	result, err = assignIndexes(a, node)
 	require.NoError(err)
 
-	lookupIdxs, ok = result["t1"]
+	_, ok = result["t1"]
 	require.False(ok)
 
 	node = plan.NewProject(
@@ -183,7 +183,7 @@ func TestAssignIndexes(t *testing.T) {
 	result, err = assignIndexes(a, node)
 	require.NoError(err)
 
-	lookupIdxs, ok = result["t1"]
+	_, ok = result["t1"]
 	require.False(ok)
 }
 
--- a/sql/analyzer/assign_indexes_test.go
+++ b/sql/analyzer/assign_indexes_test.go
@@ -1,8 +1,6 @@
 package analyzer
 
 import (
-	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/src-d/go-mysql-server/memory"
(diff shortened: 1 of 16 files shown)


sql/analyzer/assign_indexes_test.go
```diff
@@ -125,39 +135,135 @@ func TestAssignIndexes(t *testing.T) {
...
+				),
+			),
+			plan.NewResolvedTable(t1),
+		),
+	)
+
+	result, err = assignIndexes(a, node)
+	require.NoError(err)
+
+	lookupIdxs, ok = result["t1"]
```
golangcibot: SA4006: this value of `lookupIdxs` is never used (from `staticcheck`)

sql/analyzer/assign_indexes_test.go
```diff
@@ -125,39 +135,135 @@ func TestAssignIndexes(t *testing.T) {
...
+				expression.NewTuple(expression.NewLiteral(int64(1), sql.Int64), expression.NewLiteral(int64(2), sql.Int64)),
+			),
+			plan.NewResolvedTable(t1),
+		),
+	)
+
+	result, err = assignIndexes(a, node)
+	require.NoError(err)
+
+	lookupIdxs, ok = result["t1"]
```
golangcibot: SA4006: this value of `lookupIdxs` is never used (from `staticcheck`)

zachmu:
Closing in favor of https://github.com/src-d/go-mysql-server/pull/865 -- we have continued development on this branch, and this PR is big enough without adding additional features.

